PR #1, opened 2025-08-18
Implicit type casting of serializable enums

shyukahn:
This PR adds support of implicit casts of serializable enums to their underlying types within comparison operators and a minor fix regarding serializable enums with overlapping element names.

1. Implicit casts of serializable enums
    Serializable enums can be compared with equality (`==`) and inequality (`!=`) operators with (1) another serializable enum with the same underlying type and (2) arbitrary-precision integers.
    - `$reduce_senum_typ(typ)`
        This function reduces serializable enum types to their underlying types, and also applies it recursively to tuple expressions (`SeqT` in SpecTec).
    - More definitions of function `$coerce_binary`
        If the types of the two expressions are not equal nor a subtype of one another, their types are reduced with `$reduce_senum_typ` and `$coerce_binary` retries to match their types.

2. Serializable enums with overlapping element names
    Fixes an error that occured in type checking when two different serializable enums have an element whose names overlap. The type checker now classifies `p4c/testdata/p4_16_samples/dash/dash-pipeline-pna-dpdk.p4` as a well-typed program.


jaehyun1ee:
Reflects https://github.com/p4lang/p4-spec/pull/1341.

PR #2, opened 2025-08-19
Allow implicit casts for directionless arguments

shyukahn:
This PR allows implicit casts of directionless arguments, as described in the most recent version of the [P4 specification](https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-invocations). Also, it fixes a bug that allowed calling functions using named arguments but erroneously omitting parameters without a default argument.

1. Implicit casts of directionless arguments
    Previously, the types of directionless arguments were required to match with the types of the parameters. Following the P4 specification, directionless arguments are allowed to be implicitly casted to the parameter type. 
    
    - In `rule Call_convention_expr_ok/empty-not-action`, instead of checking the types of the parameter and the given argument, the argument is implicitly casted to the parameter’s type using `$coerce_unary()`.
    - `excludes/exclude-spec-discussion/impl-cast-directionless.exclude` is removed.
    - `p4c/testdata/p4_16_samples/omec/up4.p4` is added to `excludes/exclude-target-specific/selector-match.exclude`.
    - `p4c/testdata/p4_16_errors/named-arg1.p4` is added to `excludes/exclude-target-specific/annotation.exclude`.
2. Functions with named arguments and default arguments
    When calling a function with specifying each parameter’s name, parameters with default values are allowed to be omitted. However, there was a bug that allowed omitting parameters **without** default values. This is fixed by checking whether each omitted parameter has a default value in the function definition.
    
    - `$get_missing_parameters()`, which is called in `$find_match_named_default()`, is renamed to `$get_missing_default_parameters()` and now only returns missing parameters that have a default value.
    - `p4c/testdata/p4_16_samples/named-arg1.p4`, which was previously in `excludes/exclude-spec-discussion/impl-cast-directionless.exclude`, now passes type checking.


shyukahn:
I checked that the type checker rejects issue1334.p4. I added it to [overload-package](https://github.com/kaist-plrg/p4-spectec/blob/main/excludes/exclude-p4c-discussion/overload-package.exclude) in [17d35aa](https://github.com/kaist-plrg/p4-spectec/pull/2/commits/17d35aacafcfad8f14256bd775aa8a9036a0266d)

spec-concrete/5.15.1-typing-call-convention.watsup
```diff
@@ -117,8 +117,9 @@ rule Call_convention_expr_ok/empty-not-action:
   p TC NOACTION |- (`EMPTY typeIR_param _ _) `@ typedExpressionIR_arg
                  : typedExpressionIR_arg
```
jaehyun1ee, line 118: Shouldn't this be `typedExpressionIR_arg_cast`?

shyukahn: Yes, `Call_convention_expr_ok` should return the casted expression. Fixed it in [0272cd1](https://github.com/kaist-plrg/p4-spectec/pull/2/commits/0272cd111baf3a7bc22030b66281971b335de6ff).

jaehyun1ee:
Sorry for the delayed merge, now merging!

jaehyun1ee (approved):

jaehyun1ee:
Reflects https://github.com/p4lang/p4-spec/pull/1341.

PR #3, opened 2025-08-19
Compile-time evaluation of field and index accesses

shyukahn:
This PR expands (local) compile-time known values with field and index accesses. Also, a boolean value which represents the valid bit is added to header values. 
1. Compile-time known-ness of field and index accesses
    Values obtained by accessing a field of or indexing another value was previously considered dynamic (i.e. not compile-time known). Now following the P4 specification, the following expressions are considered (local) compile-time known and may be statically evaluated:
    - Accessing a field of a (local) compile-time known struct, header, or header union value.
    - Indexing a (local) compile-time known tuple or header stack value.

    With the above changes, `excludes/exclude-spec-discussion/compile-time-known-access.exclude` is removed.

    Some other minor fixes were involved:
    - `$joins_ctk(eps)` evaluates to `LCTK`.
    - Invalid header expressions(`{#}`) may be casted to another header type during static evaluation.
2. Valid bit of header values
    A boolean value is added to the syntax of a header value. It represents the hidden valid bit.
    ```
    syntax headerValue = HEADER tid `{ bool fieldValue* }
    ```
    Following the P4 specification, initialized header values are valid and the default value for header values is invalid.


jaehyun1ee (approved):

jaehyun1ee:
Reflects https://github.com/p4lang/p4-spec/pull/1342.

PR #4, opened 2025-08-19
Allow extern, parser, control, and package types to be aliased by typedef declarations

shyukahn:
This PR allows extern, parser, control, and package types to be aliased by typedef declarations.
1. `$nestable'_typedef()`
    `$nestable'_typedef` is updated to return true for `extern`, `parser`, `control`, and `package` types.
2. Constructors for aliases of `extern` and `package` types
    When `extern` or `package` types are declared, their names are used as constructors in instantiations. Aliases of those types introduced by `typedef` declarations may also be used as constructors. `typedef` declarations of `extern` and `package` types are updated to also add new constructors with the new names.
    - `rule Decl_ok/typeDeclaration-typedefDeclaration-typedef-type-constructor` is similar to `rule Decl_ok/typeDeclaration-typedefDeclaration-typedef-type`, except that it adds new constructors to the typing context.
    - `$add_typedef_constructors()` finds the constructors of the base type of the `typedef` declaration and copies the information to add the corresponding constructor with the new name.

    With the above changes, `excludes/exclude-spec-discussion/nest-typedef-object.exclude` is removed. A new type name introduced by a `typedef` declaration for an `extern` or `package` type can be used as a constructor:
    ```
    extern MyCounter<I> {
        MyCounter(bit<32> size);
    }
    
    typedef bit<10> my_counter_index_t;
    typedef MyCounter<my_counter_index_t> my_counter_t;
    
    control Test() {
        my_counter_t(1024) counter_set;
    		...
    }
    ```
3. Limitations
    Type checking fails when a type that consists of a don’t care type (”`_`”) as a type argument appears as a base type in a `typedef` declaration. For example, this `typedef` declaration is currently not allowed:
    ```
    package Switch<T>(Ingress<T, _, _> ingress, Egress<T, _, _> egress);
    typedef Switch<bit<32>> my_switch_t;
    ```


jaehyun1ee:
```p4
extern Ext<T> {
  Ext(T t);
  void boo(in T t);
}

typedef Ext<bit<32>> myExt;

myExt(32w1) myE;
```

This doesn't seem to work for the above case. Seems like the current spec doesn't take specialization of constructor itself when making an alias. i.e., `typedef Ext<bit<32>> myExt` should have as constructor type: `(bit<32> t) -> ... the extern type Ext ...`, but the current spec adds `(T t) -> ... the extern type Ext ...`, where `T` should have been specialized to `bit<32>`.

shyukahn:
Thanks for pointing that out! I made a commit that updates `$add_typedef_constructors` and `$get_typedef_constructors` to specialize constructors when creating an alias.

jaehyun1ee:
Closing in favor of #35.

PR #5, opened 2025-08-19
Instantiation of objects with abstract methods

shyukahn:
This PR disallows overloading of abstract methods and updates the typing rules for instantiating objects with abstract methods.

1. Disallow overloading of abstract methods
    
    Previously, overloading of abstract methods in an extern object declaration was allowed as long as they were distinguishable. The usage of `$add_routine_overload()` in `rule ExternMethod_ok/abstract` has been replaced with `$add_routine_non_overload()`.
    
2. Match abstract methods by method name, arity, types, and parameter directions
    
    When instantiating an extern object with abstract methods using an initializer, the provided implementations were previously matched to the declarations by comparing method names, parameter names, and type equivalence. This behavior has been updated as follows:
    
    - Method name and type equivalence is still checked.
    - Parameter names no longer need to match.
    - Parameter directions must match.
    
    Additionally, for each paramter, neither the presence of a default value nor its default value (if it exists) is currently considered in matching. This behavior is subject to change in the future.
    
    These changes are implemented in `$subst_rdenv()`, using new functions:
    
    - `$find_rdenv_by_name()`, `$update_rdenv_by_name()`: The type `rdenv` is a map whose type of the keys is `rid`, consisting of the function/method name and its parameters. Since parameter names are ignored, general map functions such as `$find_map()` and `$update_map()` are unsuitable.  These new functions match only on method names, not parameter names.
    - `$check_abstract_method_impl()`: This function checks type equivalence and parameter directions.
    
    `excludes/exclude-spec-discussion/overload-abstract-method.exclude` is removed, following the changes.


jaehyun1ee:
Closing in favor of #36.

PR #6, opened 2025-08-19
Compound assignments

shyukahn:
This PR adds support for compound assignments in P4.

1. `rule Stmt_ok/assignmentStatement-compound`
    
    A new rule is added to the `Stmt_ok`  relation to support  compound assignments such as `+=`, `*=`, etc. For example, as `A += B;` is equivalent to `A = A + B;`, after converting the lvalue `A` to an expression, the binary operation `A + B` is type checked by `Expr_ok`. Two helper functions are used:
    
    - `$lvalue_to_expression()`: `$lvalue()` was moved to `5.08-typing-lvalue.watsup` and renamed.
    - `$assignop_to_binop()`: Converts an `assignop` to the corresponding `binop` . For example, ```$assignop_to_binop(`+=)``` returns ``` `+ ```.
2. Changes in `exclude` 
    - `excludes/exclude-future/opassign.exclude` has been removed.
    - `p4c/testdata/p4_16_samples/opassign1.p4` has been moved to `excludes/exclude-spec-discussion/div-mod-fixed.exclude` because it uses `%=` with fixed-width unsigned integers.


jaehyun1ee (approved):
LGTM! Added dynamic evaluation rules for compound assignments as well.

jaehyun1ee:
Reflects https://github.com/p4lang/p4-spec/pull/1359.

PR #8, opened 2025-09-15
Update negative type checker test generator backend and spec

jaehyun1ee:
This PR updates the negative type checker test generator backend and also the spec.

**(1) Negative type checker test generator backend**

We updated the negative type checker test generator backend. Not much changes other than misc bugfixes, but has one change to the SL AST. Instead of `HoldE` as an expression, we now have `HoldI` and `NotHoldI` as instructions.

`HoldE of id * notexp` used to represent "when a relation named `id` given inputs `notexp` holds." These apply for relations with 0 output, such as `Type_alpha` and `Sub_expl/impl` relations.  At the source (EL) level, they are represented as premises, e.g.,

```
def $coerce_binary(typedExpressionIR_l, typedExpressionIR_r)
  = (typedExpressionIR_l, typedExpressionIR_r)
  -- if _ `( typeIR_l _ ) = typedExpressionIR_l
  -- if _ `( typeIR_r _ ) = typedExpressionIR_r
  -- Type_alpha: typeIR_l ~~ typeIR_r

def $coerce_binary(typedExpressionIR_l, typedExpressionIR_r)
  = (typedExpressionIR_l_cast, typedExpressionIR_r)
  -- if _ `( typeIR_l ctk_l ) = typedExpressionIR_l
  -- if _ `( typeIR_r _ ) = typedExpressionIR_r
  -- Type_alpha:/ typeIR_l ~~ typeIR_r
  -- Sub_impl: typeIR_l <: typeIR_r
  -- if typedExpressionIR_l_cast
      = (CastE typeIR_r typedExpressionIR_l) `( typeIR_r ctk_l )
```

`Type_alpha:` to express that it holds, and `Type_alpha:/` (notice the slash after colon) to express that is does not hold.

We found it more useful to represent them as instructions in SL, so instead of having `HoldE` we have `HoldI` and `NotHoldI` instead. Their interpretations are more or less the same.

**(2) Updated spec**

We updated the spec to better resemble the P4 grammar in the official specification. The original spec is now under `spec-old` and the new spec is under `spec`.

One change is that the P4 program frontend of P4-SpecTec has also changed to handle the updated grammar. We've adjusted the P4 parser in P4-SpecTec to be compatible with the updated grammar. That said, the parser is no longer compatible with the old grammar.

Yet, the rest (value AST and interpreter logic) remain unchanged.


jaehyun1ee:
@cfbolz A heads-up before merging some changes to P4-SpecTec and the mechanized spec :)

PR #9, opened 2025-09-15
feat: command for exporting P4 program value to a JSON string

cfbolz:
add another toplevel command that exports a P4 program value to a JSON string


jaehyun1ee:
LGTM! Thanks

PR #11, opened 2025-09-16
fix: use head of list as fold_left start value in min

### LINKED ISSUE #10: 'suspicious' code in min builtin
While porting min and max I noticed that there's some logic in `min` that looks wrong to me:

```ocaml
let min (at : region) (targs : targ list) (values_input : value list) : value =
  Extract.zero at targs;
  let values =
    Extract.one at values_input |> Value.get_list |> List.map bigint_of_value
  in
  let min = List.fold_left Bigint.min Bigint.zero values in
  value_of_bigint min
```

The fact that `fold_left` starts with `0` means that `min([1, 2, 3, 4])` would be evaluated to `0`, no?

In my Python version I use the first element as the initial value for the fold, should I try to open a PR with that?

cfbolz:
Is there a test you would have wanted for this? Also, should the empty-list case error or keep returning 0?

fixes #10

### DIFF
--- a/p4spec/bin/main.ml
+++ b/p4spec/bin/main.ml
@@ -415,18 +415,21 @@ let json_ast_command =
                Format.printf "Error while parsing %s: %s" filename err))
 
 let p4_program_value_json_command =
-  Core.Command.basic ~summary:"convert a P4 program to a value and output as JSON"
+  Core.Command.basic
+    ~summary:"convert a P4 program to a value and output as JSON"
     (let open Core.Command.Let_syntax in
      let open Core.Command.Param in
      let%map includes_p4 = flag "-i" (listed string) ~doc:"p4 include paths"
      and filename_p4 = flag "-p" (required string) ~doc:"p4 file to convert" in
      fun () ->
        try
-         let value_program = Interp_sl.Typing.convert_program_to_value includes_p4 filename_p4 in
+         let value_program =
+           Interp_sl.Typing.convert_program_to_value includes_p4 filename_p4
+         in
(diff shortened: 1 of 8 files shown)


jaehyun1ee:
Thank you for the fix!

We don't have a good use of `$min` by far, and I think it would be more helpful to raise an error here (using `interp-sl/error.ml`).

We do have a single use of `$min` that implements the [`minSizeInBits` builtin method](https://p4.org/wp-content/uploads/sites/53/2024/10/P4-16-spec-v1.2.5.html#:~:text=field%20f2%200-,9.%E2%80%82Compile%2Dtime%20size%20determination,-The%20method%20calls).

```
def $sizeof_minSizeInBits''(HeaderUnionT _ (typeIR _)*)
  = $min($sizeof_minSizeInBits'(typeIR)*)
```

For this, I think we might want to change this also to:

```
def $sizeof_minSizeInBits''(HeaderUnionT _ (typeIR _)*)
  = 0
  -- if |typeIR*| = 0
def $sizeof_minSizeInBits''(HeaderUnionT _ (typeIR _)*)
  = $min($sizeof_minSizeInBits'(typeIR)*)
  -- if |typeIR*| =/= 0
```

Checking whether it passes all tests in `make test-spec` would be enough, I think.

jaehyun1ee (changes requested):
Thank you for making the changes :) As mentioned in the comments, can you try merging `main` into your fork, to address some spurious additions?

cfbolz:
closing, I'll reopen a new one without the p4c changes

PR #14, opened 2025-09-25
Enhanced failtrace outputs

pacokwon:
This branch aims to enhance the current failtrace outputs in the IL interpreter.

This approach applies domain-specific heuristics based on... lots of observation and testing 😆 The following are the main methods used in this approach.

## 1. Attach a `reason` label to every failure node
At a point of failure, a reasonable guess is made to indicate the cause of the failure. In each failtrace node, its reason for failure is attached.
```ocaml
type reason =
  | RootClause of int * int (* clause idx, max (premise_idx) *)
  | MismatchClause of int * int (* clause idx, max (premise_idx) *)
  | Root of int (* premise idx *)
  | Mismatch of int (* premise idx *)
  | Unknown

type failtrace = Failtrace of region * string * reason * failtrace list

```
## 2. Define and detect common "guard patterns" to identify root causes for each failure node
When an invalid statement is checked against a relation such as `Stmt_ok`, it fails against every rule in the relation like this:
```plaintext
- because condition statement <: emptyStatement was not met (spec-concrete/1-syntax.watsup:613.25-613.27)
- because condition statement <: assignmentStatement was not met (spec-concrete/1-syntax.watsup:624.57-624.59)
- because condition statement <: callStatement was not met (spec-concrete/1-syntax.watsup:631.30-42.45)
- because condition statement <: callStatement was not met (spec-concrete/1-syntax.watsup:632.52-63.67)
- because condition statement <: directApplicationStatement was not met (spec-concrete/1-syntax.watsup:638.47-99.51)
- because condition statement <: returnStatement was not met (spec-concrete/1-syntax.watsup:645.5-645.14)
- because condition statement <: returnStatement was not met (spec-concrete/1-syntax.watsup:646.5-646.25)
- because condition statement <: exitStatement was not met (spec-concrete/1-syntax.watsup:652.24-652.31)
- because condition statement <: blockStatement was not met (spec-concrete/5.10-typing-statement.watsup:155.21-155.35)
- because condition statement <: conditionalStatement was not met (spec-concrete/1-syntax.watsup:669.5-165.36)
- because condition statement <: conditionalStatement was not met (spec-concrete/1-syntax.watsup:670.5-178.36)
- because condition statement <: switchStatement was not met (spec-concrete/1-syntax.watsup:735.26-278.66)
- because condition statement <: switchStatement was not met (spec-concrete/1-syntax.watsup:735.26-382.66)
```
In this case, these subtype expressions are the first premises to appear in every rule. "guard patterns" like these are hand written and matched against failtraces to see if a rule is failing simply because it doesn't belong there. We mark their reasons as `Mismatch`.

## 3. Compare failure points within rules and favor those that progress "deeper"
A problem that arises from only using method 2 is that some rules are inevitably marked as "Mismatch"es even when they are legitimate causes of failure. This happens because they "look" they belong to a mismatch pattern but they are actually not. To complement this problem, we also compare how "deep" the program had to go through in a given rule. 

An example is [`typecheck_e.p4`](https://github.com/p4lang/p4c/blob/b068f6d4e7430c643b71c2097bed37e4461419e7/testdata/p4_16_errors/typecheck_e.p4), one of the negative tests in `p4c`. A naive failtrace of this program, at the deepest level, looks like this:
```plaintext
1. because application of rule Decl_ok/constantDeclaration failed (spec-concrete/5.11-typing-declaration.watsup:953.6-953.13)
  - because condition declaration <: constantDeclaration was not met (spec-concrete/1-syntax.watsup:774.18-774.48)
...
13. because application of rule Decl_ok/controlDeclaration failed (spec-concrete/5.11-typing-declaration.watsup:953.6-953.13)
  - because condition typeParameterListOpt matches ``EMPTY` was not met (spec-concrete/1-syntax.watsup:184.5-184.11)
...
25. because application of rule Decl_ok/typeDeclaration-packageTypeDeclaration failed (spec-concrete/5.11-typing-declaration.watsup:953.6-953.13)
  - because condition declaration <: packageTypeDeclaration was not met (spec-concrete/1-syntax.watsup:1197.18-908.85)
```

While clause 13 is the real cause of the program failing, it is detected as a "Mismatch" because looks as if it is using a guard pattern that other clauses actually use, although that is not actually the case. One thing that differentiates clause 13 from all other clauses is that `typecheck_e.p4` progresses through clause 13 deeper than it does for any other clauses in the relation. We make use of this observation and identify clause 13 as the root cause of the program.

---

This branch also adds the `--full-trace` flag to `run-il`. When the flag is enabled, the full failtrace is printed out whenever a p4 program fails. By default, this is turned off.

## TODO
- [x] Produce a working version of the enhanced failtrace heuristic for 20 pre-selected negative tests
- [x] Verify performance for the rest of the negative tests


pacokwon:
Does `--full-trace` look okay for the name of the flag? I'd like to follow the "single dash" convention but `-full-trace` looked odd to me. Suggestions would be appreciated!

p4spec/.ignore
jaehyun1ee: Please remove this file.

p4spec/lib/interp-il/trace_mode.ml
```diff
@@ -0,0 +1,5 @@
+type mode = [ `Full | `Concise ]
```
jaehyun1ee: Is there a specific reason for making it a polymorphic variant? I would prefer making it `type mode = Full | Concise`.

p4spec/lib/interp-il/trace.ml
```diff
@@ -286,3 +286,62 @@ let profile (trace : t) : unit =
...
+type subexp' = Let of exp' * exp' | If of exp'
+
+let try_extract_exp = function
+  | Prem p -> (
+      match p.it with
+      | IfPr ifpr -> Some ifpr.it
+      | LetPr (_, ({ it = DownCastE _; _ } as rvalue)) -> Some rvalue.it
+      | LetPr (({ it = CaseE _; _ } as lvalue), _) -> Some lvalue.it
+      | _ -> None)
+  | _ -> None
```
jaehyun1ee: Isn't this deadcode?

p4spec/lib/util/attempt.ml
```diff
@@ -26,22 +45,45 @@ let rec choice = function
...
+         | _, Root premise | _, RootClause (_, premise) ->
+             RootClause (i, premise)
+         | MismatchClause (_, premise_1), MismatchClause (_, premise_2) ->
+             if premise_1 < premise_2 then MismatchClause (i, premise_2)
+             else acc
+         | _, MismatchClause (_, max_premise) | _, Mismatch max_premise ->
+             MismatchClause (i, max_premise)
+         | _, Unknown -> acc)
+       Unknown
+
```
jaehyun1ee: Nitpicking, but I prefer using `_a`, `_b` over `_1`, `_2`.

jaehyun1ee (changes requested):
Thank you, the failtrace surely looks nicer with the updates. I made some comments, mostly nitpicking. Overall, it would be nice to have a brief explainer of the trace-pruning heuristics in the code as comments.

p4spec/.ignore
pacokwon: Eliminated in [2d99816](https://github.com/kaist-plrg/p4-spectec/pull/14/commits/2d99816f8150ceea21d64fe28819404f1b322af7).

PR #15, opened 2025-09-26
use head of list as fold_left start value in min and forbid empty lists as arguments to min and max

### LINKED ISSUE #10: 'suspicious' code in min builtin

cfbolz:
fixes #10. it's a re-do of #11, which had unrelated changes.

### DIFF
--- a/p4spec/lib/interp-il/builtin/nats.ml
+++ b/p4spec/lib/interp-il/builtin/nats.ml
@@ -2,6 +2,7 @@ open Xl
 open Il.Ast
 module Value = Runtime_dynamic.Value
 open Util.Source
+open Error
 
 (* Conversion between meta-numerics and OCaml numerics *)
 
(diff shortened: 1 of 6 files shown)


jaehyun1ee:
Thank you for the PR, merging!

PR #16, opened 2025-09-26
feat: add github ci workflow for pull requests and pushes to main

cfbolz:
it installs ocaml, builds p4spectec, runs the tests, generates an json file from the spec ast and validates that its correct json.


jaehyun1ee:
It is very nice to have a CI running on our project :) Thank you!

PR #19, opened 2025-10-01
fix: make use p4program VarT, like the spec defines

cfbolz:
I'm a little bit unsure about this PR, but ran into the following issue today: the 'new' spec uses the type `p4program` for the P4 program values:

https://github.com/kaist-plrg/p4-spectec/blob/48fcc27d3f305c2789179c5aed3959cf1b967eba/spec/1-syntax.watsup#L690

the old spec used `program`:

https://github.com/kaist-plrg/p4-spectec/blob/48fcc27d3f305c2789179c5aed3959cf1b967eba/spec-old/1a-syntax-el.watsup#L325

This PR makes the parser produce a `p4program` `VarT`.

### DIFF
--- a/p4spec/lib/interface/parser.mly
+++ b/p4spec/lib/interface/parser.mly
@@ -1564,4 +1564,4 @@ declarationList_:
   | dl = declarationList_ { dl |> wrap_list_v "declaration" }
 
 p4program:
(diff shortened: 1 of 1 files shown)


jaehyun1ee:
Thanks for the catch!

PR #21, opened 2025-11-07
feat: implement Lvalue_write for header stack

pacokwon:
`Lvalue_write/stack` implements lvalue write semantics for header stacks.

One thing I'm iffy with is `$update_headerStack` which is basically an application of map function with indices over a sequence. Suggestions would be appreciated!


spec-concrete/8.05-eval-lvalue.watsup
```diff
@@ -112,6 +112,18 @@ rule Lvalue_write/referenceExpressionIR:
...
 rulegroup Lvalue_write/storageReference-nameIR {
 
+  rule Lvalue_write/stack:
+    p EC_0 STO |- (storageReference `. "next") -> value : EC_1
+    ---- ;; read base value
+    -- Lvalue_read:
+        p EC_0 STO |- storageReference : value_base
+    -- if HEADER_STACK `[ value_element* `( n_idx; n_size ) ] = value_base
+    -- if value_element_update*
+        = $update_headerStack(value_element*, value, n_idx, n_size)
```
jaehyun1ee: `    -- if value_element_update* = value_element* [ [n_idx] = value ]`

This may be a better version, which uses `UpdE`. But I think the IL/SL interpreter for this particular feature is not yet implemented. If it turns out to be the case, let's just use `$update_headerStack` for now.

jaehyun1ee:
LGTM, merging!

PR #30, opened 2025-11-14
Add an Emacs major mode for *.watsup files

3tty0n:
Try to add an Emacs major mode for watsup. Please suggest a suitable name for this major mode.

You can check how it works in Emacs by loading `M-x load-file editor/emacs/watsup-mode.el` and opening *.watsup files.

### DIFF
--- a/editor/emacs/watsup-mode.el
+++ b/editor/emacs/watsup-mode.el
@@ -1,0 +1,88 @@
+;;; watsup-mode.el --- Major mode for Watsup specification language -*- lexical-binding: t; -*-
+;;; Code:
+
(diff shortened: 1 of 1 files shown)


jaehyun1ee:
@3tty0n Oops, seems like I accidentally assigned Copilot for review. Actually, I'm not used to Emacs, so I don't think I have enough knowledge to review this PR. I'll merge this PR for now, and revisit if we have more Emacs users :)

3tty0n:
@jaehyun1ee Oh, thanks! Tbh I'm a heavy user of Emacs :joy:

PR #34, opened 2025-12-02
feat: Table functions

KunJeong:
This PR extends P4-SpecTec with a new feature: table functions.

## Role and Syntax
A table function is a new syntactic construct for rendering functions into an asciidoc table.
Its main use-cases are:
`$is_[...]able_typeIR` functions in `2.6-type-aux-property.watsup`
`$nestable'_[...]` functions in  `5.03-typing-wellformed.watsup`
`$compat_[...]` functions in `5.06.1-typing-expression.watsup`

Because not all functions can be rendered as a table, we would like to encode such restrictions into the syntax. Thus, we introduce new syntax instead of using function hints.

Below is an example of a table function.
```spectec
tbl dec $nestable'_header(typeIR) : bool
tbl def $nestable'_header =
    | BOOL => true
    | INT `< _ > => true
    | BIT `< _ > => true
    | VARBIT `< _ > => true
    | nameTypeIR => true
    | newTypeIR => true
    | ENUM _ `< _ > `{ _ } => true
    | STRUCT _ `{ (typeIR_field _ `;)* }
        => $for_all($nestable_struct_in_header(typeIR_field)*)
    | _ => false
```

Unlike normal functions (renamed *"plain functions"* for distinction) that can have their definitions split into multiple clauses, a table function has a single definition point, and a pattern-matching style syntax that distinguishes it from plain functions.
There are several constraints enforced in the elaboration stage.
1. A table function must return a boolean expression.
2. Each *"table-match"* (the LHS of each row) only allow *"shallow bindings"*. Shallow bindings are explained further in a later section.
3. The set of table-matches must be mutually exclusive and form a complete function. We allow using the `_` (wildcard) pattern as the final pattern for convenience. We will refer to the closing wildcard pattern as a *"closer"*.

Collectively, these constraints guarantee that the function can be rendered in table form.

There is also experimental support for table functions with two or more parameters. This is what the syntax looks like:
```spectec
tbl dec $compat'_satplusminus(typeIR, typeIR) : bool

tbl def $compat'_satplusminus =
  | (INT `< w_l >, INT `< w_r >) => $(w_l = w_r)
  | (BIT `< w_l >, BIT `< w_r >) => $(w_l = w_r)
  | (_, _) => false
```
We allow the use of tuple expressions to match against multiple arguments.

The abbreviation `tbl` is to prevent collision with the "table" in some rule names such as `CallableType_ok/typedExpressionIR-nameIR-table-apply-method`.
## Overview
Data representation
```ocaml
(* el/ast.ml *)
| TableDecD of id * param list * plaintyp * hint list
| TableDefD of id * (exp * exp) list
```
On the EL level, table declarations are declarations minus type parameters.
At this stage, the table definition body is just a list of expression pairs (pattern, output).
Note that the function parameters are anonymous, similar to the OCaml `function` construct.
```ocaml
(* il/ast.ml *)
type tablerow = tablerow' phrase
and tablerow' = exp list * arg list * exp * prem list

| TableDecD of id * param list * typ * tablerow list * hint list
```
After elaboration, each row of the table (`tablerow`) is a list of input patterns (the row *"signature"*), a list of arguments, the output expression and the generated premises. The signature is only for rendering purposes, and interpretation is done using the arguments and the premises. As a `tablerow` is basically a clause with a signature attached, executing a table function is nearly identical to executing a normal function.
```ocaml
(* sl/ast.ml *)
type tablerow = exp list * exp * instr list
type tablefunc = id * arg list * typ * tablerow list * hint list

| TableDecD of tablefunc
```
Structuring leaves the signature and the return expression untouched, as we need this for table rendering. The generated instruction list still contains a `ReturnI` at the appropriate spot; this is a deliberate duplication of data. The tablerow arguments are antiunified into the function definition, so we no longer have to bind arguments for each row.

When executing a table function on SL, we first find the matching row using the row signatures which is verified to succeed by the elaborator. Then we execute the row normally.

The AST in the prose generation stage is identical to that of SL.

## Elaboration
Elaboration of a table function declaration checks that the return type is a boolean and adds the declaration to the environment.

Elaboration of a table function definition is done in the following steps, generally similar to plain functions. The differences are highlighted in the next section.

- Elaboration of a `tablerow`:
    1. each EL `tablerow` is a pair `(exp_pat, exp_body)`
    2. If `exp_pat` is a tuple, each sub-expression becomes an argument. **This adds support for multi-parameter tables.**
    3. The arguments are elaborated into `args_il`
    4. A local context is initialized and the set of free identifiers collected.
    5. Dataflow-analysis on `args_il` as binding, which also generates implicit premises
    6. Dataflow-analysis on `args_il` as bound, which generates the signatures.
    7. Elaboration and dataflow-analysis on `exp_body`, reusing helper function for plain function clauses.
- A validity check for the list of tablerows, which checks whether each case of the pattern match is exclusive and the list is exhaustive.
### Shallow binding analysis
The idea of a "shallow" binding is one step of variant destructuring.
We want to allow patterns like:
- `newTypeIR`: simple downcasting
- `` ENUM _ `< _ > `{ _ } ``: simple pattern-matching
- `` STRUCT _ `{ (typeIR_field _ `;)* } ``: simple pattern-matching + type alias destructuring

We do not want to allow things like:
- `` TYPEDEF _ namedTypeIR ``: additionally constrains the inner `typeIR` to `namedTypeIR`
- `` (INT `< w >, INT `< w >) ``: additionally constrains the two widths to be equal

Thus, the dataflow analysis in steps 5 and 6 have an additional check built-in, implemented in `dataflow/shallowbind.ml`. The shallowbind check precedes the existing `Collectbind`, `Multibind` and `Partialbind` and performs a very simple pattern matching to check that the argument is indeed "shallow." There is second check after the `Multibind` pass that confirms no side-conditions were generated.
### Check exclusiveness/exhaustiveness
After all rows are elaborated, we perform checks on the set of match patterns as a whole. These checks are generalized for tuples. We have two types in `elaborate/pattern.ml`: `PatternSet.t` which is a set of `nottyp`s covered by one expression, and `tuple_pattern` which represents what patterns are matched by a tuple.

Two helper functions `patterns_covered_by_exp` and `patterns_covered_by_typ` retrieve the `PatternSet` for an expression and a typ (`VarT`), respectively; these reside in `elab.ml` because they must query the typedef environment and elaborate `El.Ast.NotationT` into `Il.Ast.nottyp`.

The main logic `check_valid_match_tablerows` first extracts the *"closer"* (terminating all-wildcard pattern) if there is one, which may let us skip the exhaustiveness check.

Internally, the exclusiveness check simply checks if two `tuple_pattern`s overlap, and the exhaustiveness check repetitively subtracts the current row (also a `tuple_pattern`) from the "total" pattern resulting in the set of missing patterns, which we can print out to users.
This pattern-matching check is very general, and has the potential to be later used in non-table functions.
## Next steps
This PR only implements the core functionality of table functions, and leaves the following questions open for future improvement.
#### Styling
Tables are rendered in asciidoc with the default styling with no options passed in. However, one would want to be able to customize the appearance of tables, as well as texts in header rows and columns, to a certain degree. What customizations we would allow, and what kind of hints would be necessary to support such customizations require further discussion.
#### 2D-rendering of 2-param table functions
We haven't yet discussed whether we want to enable rendering two-parameter table functions as two-dimensional tables. The notion of multi-param table functions in general is underdeveloped at the moment.
### Table groups
Table groups are another syntactic construct that *groups* multiple table functions into a single 2D table.
For example, one could merge `nestable'_header`, `nestable'_struct`, `nestable'_list`, etc. into a single 2D table where each column is the container kind. This more closely mimics what we already have in the P4 specification.
```adoc
[.center,cols="3,^2,^2,^2,^2",width=80%]
|===
| Element type 6+^| Container kind  
|                | header    | header_union | struct or tuple | list | array [4] | header stack
| `bit<W>`       | allowed   | error   |  allowed | allowed | allowed | error
| `int<W>`       | allowed   | error   |  allowed | allowed | allowed | error
| `varbit<W>`    | allowed   | error   |  allowed | allowed | allowed | error
...
|===
```
For this purpose, a table group can be defined like the following:
```spectec
tblgroup nestable'(typeIR) : bool {
    tbl dec $header
    tbl dec $struct_or_tuple    hint(tbl_col "struct or tuple")
    ...
}

tbl def $nestable'.header =
    | BOOL => true
    | INT `< _ > => true
    | BIT `< _ > => true
    ;; ...
    | _ => false
```
This keeps multiple table functions with the same signature in one group function, and as a bonus we can call the inner functions with a dot-access notation.

All semantic restrictions are already built into the table function elaborator, so we only need the following:
1) A construct `tablegroup` in EL/IL/SL/PL, and parser support.
2) Modifications to function environment and function invocation to support dot-accesses on tablegroups.
3) Prose support to render a tablegroup into a 2D table, and propagation of necessary hints.
4) As a part of 3), a pass that can compute the smallest subset of patterns needed to incorporate table functions with different pattern granularity. This is perhaps the most challenging part of table groups


jaehyun1ee (approved):
LGTM! Thank you for the commits, I added some refactorings, mostly renaming types, variables, and functions.

And one change to the SL interpreter, where the previous version used the table row signatures to first match a table row and then apply the instructions, the updated version instead concatenates all instructions from each table rows and runs them as a whole (without using the signatures). Please take a look at the [interpreter code](https://github.com/kaist-plrg/p4-spectec/blob/c4d9a8ae924bdf0b2c4347e7cd15a07df0d8c872/p4spec/lib/interp-sl/interp.ml#L1560-L1574).

Merging.

KunJeong:
Thanks for the detailed review.

Regarding the SL interpreter, the new code looks much cleaner 🤩

At the time of implementation I thought it was better to leverage the properties we already know about table functions, which ensures no backtracking when executing each row. While I couldn't come up with a clean implementation that encodes this property, I'm leaving this comment in the hopes that we could revisit this in the future.

PR #35, opened 2025-12-18
Allow extern, parser, control, and package types to be aliased by typedef declarations

jaehyun1ee:
This re-implements #4, which allows extern, parser, control, and package types to be aliased via `typedef`.

1. `$nestable'_typedef()`

`$nestable'_typedef` is updated to return true for extern, parser, control, and package types.

2. Constructors for aliased object types

The previous PR handles aliasing by creating aliased constructors when type checking `typedef` declaration.
This PR implements it the other way around, where aliased constructors are resolved when invoking a constructor, inspired by the P4C patch: https://github.com/p4lang/p4c/pull/3953.

3. Limitations on hidden type parameters

Limitations are inherited from the previous PR. Type checking fails when a type that consists of a don’t care type (”_”) as a type argument appears as a base type in a typedef declaration.

Quick idea: allow "partial" specialization for hidden type parameters,

```
rule Type_wf/specializedTypeIR:
  bound |- typeDefIR_base `< typeIR_arg* >
  -- if typeIR_spec = $specialize_typeDefIR(typeDefIR_base, typeIR_arg*)
  -- Type_wf: bound |- typeIR_spec
```

In the first premise, only specialize the explicit type parameters.


jaehyun1ee:
Reflects https://github.com/p4lang/p4-spec/pull/1343.

PR #44, opened 2025-12-19
Coercion for serializable enums

jaehyun1ee:
This PR improves https://github.com/kaist-plrg/p4-spectec/pull/1.

### Binary coercion involving serializable enums

Serializable enums are quite tricky, because it can be *implicitly* cast to its underlying type when necessary. This complicates binary coercion, because there may be the case where `x` and `y` (i) have unequal types, and (ii) neither is the subtype of the another (i.e., `x </: y` and `y <:/ x`) but the two have a common super-type.

For instance, binary coercion of `bit<8>` and `int` is easy because the rhs, `int`, can be implicitly cast to `bit<8>`. The case for `int` and `bit<8>` works vice-versa.

Now consider the case with serializable enums:

```p4
enum bit<8> E1 {
   e1 = 0, e2 = 1, e3 = 2
}

enum bit<8> E2 {
   e1 = 10, e2 = 11, e3 = 12
}

control c() {
    apply {
        E1 a = E1.e1;
        E2 b = E2.e2;
        bool x = (a == b); // HERE
   }
}
```

The `HERE` part is legal, because this can be implicitly cast to `(bit<8>) a == (bit<8> b)`. But implementing such is tricky. Notice that `a`'s type, `E1`, and `b`'s type, `E2`, are unequal and also neither is the subtype of another. The type checker should somehow figure out their common super-type, `bit<8>`.

AFAIK, serializable enums are the only type that introduces such behavior. So the mechanized spec special-cases on serializable enum types on coercion. Here is the overall process for `$coerce_binary`:

```
dec $coerce_binary(typedExpressionIR, typedExpressionIR)
  : (typedExpressionIR, typedExpressionIR)?
  hint(prose_in %0 "and" %1 "implicitly cast to equal types")
```

* If the two expressions' types are alpha-equivalent, then the result is the same.
* If the left is a subtype of the right, cast left to right.
* If the right is a subtype of left, cast right to left.
* Try reducing serializable enums on the left side, then re-try coercion.
* Try reducing serializable enums on the right side, then re-try coercion.

With the above process, `a == b` is coerced as:

* Reduce lhs to `(bit<8>) a`. And re-try coercion of `(bit<8>) a` and `b`.
* Because rhs is a subtype of lhs, cast rhs to `(bit<8>) b`.
* Now the two types are the same as: `bit<8>`.

### Special-casing key types in `select` and `table`s

Some programs use serializable enums as `select` or `table` keys, where the match-entries use the reduced type. e.g.,

```p4
E1 a = E1.e1;
transition select (a) {
    0x01: accept;
    default: getopt;
}
```

In this case, the match-entry `0x01` is expected to be *implicitly* cast to the key type, `E1`. However, such is prohibited as per the spec, since serializable enums can only be implicitly cast to the underlying type, but not vice-versa.

But to support such uses, key types in `select` and `table` are reduced to their underlying types. Thus, the above snippet is type-checked as:

```p4
E1 a = E1.e1;
transition select ((bit<8>) a) {
    0x01: accept;
    default: getopt;
}
```

Both parts are required to pass tests:

* [p4c/testdata/p4_16_samples/enumCast.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/enumCast.p4)
* [p4c/testdata/p4_16_samples/issue3333.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/issue3333.p4)
* [p4c/testdata/p4_16_samples/table-entries-ser-enum-bmv2.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/table-entries-ser-enum-bmv2.p4)

### DIFF
--- a/excludes/exclude-bug/coerce-serializable-enum.exclude
+++ b/excludes/exclude-bug/coerce-serializable-enum.exclude
@@ -1,3 +1,0 @@
-p4c/testdata/p4_16_samples/enumCast.p4
-p4c/testdata/p4_16_samples/issue3333.p4
-p4c/testdata/p4_16_samples/table-entries-ser-enum-bmv2.p4
--- a/excludes/exclude-p4c-discussion/tuple-lvalue.exclude
+++ b/excludes/exclude-p4c-discussion/tuple-lvalue.exclude
@@ -1,1 +1,0 @@
-p4c/testdata/p4_16_samples/issue4057.p4
--- a/excludes/exclude-target-specific/switch-within-action.exclude
+++ b/excludes/exclude-target-specific/switch-within-action.exclude
@@ -1,1 +1,0 @@
-p4c/testdata/p4_16_errors/issue3299.p4
--- a/p4spec/test/elab.expected
+++ b/p4spec/test/elab.expected
@@ -3303,17 +3303,25 @@ def $write_value_from_bits'(value, nat, bool*) : (value, bool*) =
   -- let b_t*{b_t <- b_t*} = b*{b <- b*}[n_var : n_t]
   -- let i = $int_to_bitstr(n_var as int, $bits_to_int_signed(b_h*{b_h <- b_h*}))
 
-  ;; ../../../../spec-concrete/2.1.2-value-aux.watsup:187.1-191.63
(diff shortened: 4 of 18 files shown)


jaehyun1ee:
@shyukahn for visibility.

PR #101, opened 2026-01-21
Organize excludes for dynamic semantics + Spec/V1Model updates

pacokwon:
## Organize Excludes
Tests in `dynamic/excludes` have been grouped by more specific causes. These are the categories of currently excluded tests:
* CRC32 (to be implemented)
* Extern function without implementation
* STF multicast instructions (`mc_mgrp_create`, etc)
* Relies on undefined behavior

## Spec
* Handle StackOutOfBounds error where it is possible
* Support `write_value_from_bits` for header stack
* Fix unnamed enum values to use originally assigned underlying value
* Handle `static_assert` during typechecking phase
* etc.

## V1Model
* Implement externs (`log_msg`, `digest`)
* Fix padding issue for checksum
* etc.

## Tests
P4Testgen-generated STF files that use `setdefault` have been patched to corretly use fully qualified names for tables.


jaehyun1ee (approved):
LGTM! Merging.

PR #102, opened 2026-01-21
Support adding table entries via STF files with keys whose match kind is lpm

shyukahn:
This PR mainly adds support of adding table entries via STF files with keys whose match kind is lpm. Additionally, if an STF file adds a table entry with a key whose match kind is ternary, the priority must be explicitly provided. Binary values in STF files are also supported.

### Match kind `lpm`

In the `add` command of STF files, the value of table keys are provided by either a single value which may have wildcard digits (`*`) or a value followed by a `/` and the number of bits, similar to the CIDR notation. To also support slash notations, `tableKeyInterface` is now a pair of a key’s name and a `tableKeyValueInterface`, which is newly defined.

```
syntax tableKeyValueInterface =
  | `NUM text
  | text `SLASH nat
syntax tableKeyInterface = (nameIR, tableKeyValueInterface)
```

The value of lpm keys may be provided by both ways. For example, consider `lpm_ebpf.p4`:

```p4
#include <ebpf_model.p4>
#include <core.p4>

#include "ebpf_headers.p4"

struct Headers_t { ... }

parser prs(packet_in p, out Headers_t headers) { ... }

control pipe(inout Headers_t headers, out bool pass)
{
    action Reject(IPv4Address add) { ... }

    table Check_src_ip {
        key = { headers.ipv4.srcAddr : lpm; }
        actions =
        {
            Reject;
            NoAction;
        }
        ...
    }

    apply {
        ...
        Check_src_ip.apply();
    }
}

ebpfFilter(prs(), pipe()) main;
```

An STF file may add an entry by the following command using the slash notation:

```
add main.filt.Check_src_ip headers.ipv4.srcAddr:0x0a01/16 Reject(add:0)
```

Equivalently, one can use a single number with wildcard digits:

```
add main.filt.Check_src_ip headers.ipv4.srcAddr:0x0a01**** Reject(add:0)
```

However, unlike ternary keys, `*` cannot precede any numerical digit when providing the value of a key. For example, the following command is not allowed since `headers.ipv4.srcAddr`'s match kind is lpm:

```
# This STF command is not allowed
add main.filt.Check_src_ip headers.ipv4.srcAddr:0x**0a01** Reject(add:0)
```

Now `lpm_ebpf.p4` passes the simulator test.

### Require explicit priority when needed

When at least one key’s match kind is ternary in a table, a priority value must be explicitly given by the control plane (STF file) when it adds an entry. `$needs_priority_tableKeysPropertyIR()` is a new helper function that checks this condition.

```
dec $needs_priority_tableKeysPropertyIR(tableKeysPropertyIR) : bool
def $needs_priority_tableKeysPropertyIR(KEY `= `{ tableKeyIR* }) = true
  -- (if _ `: nameIR_matchKind _ `; = tableKeyIR)*
  -- if $exists_((nameIR_matchKind = "ternary")*)
def $needs_priority_tableKeysPropertyIR(tableKeysPropertyIR) = false
  -- otherwise
...

def $tableObject_add_entry(
    TABLE typeId `{ frame TBL },
    tableEntryPriorityInterface,
    tableKeysetInterface,
    tableActionInterface)
  = TABLE typeId `{ frame TBL_updated }
  ---- ;; create a table entry from the interface
  -- if tableEntryPriorityOptIR
      = $tableObject_create_entry_priority(tableEntryPriorityInterface)
  -- if $needs_priority_tableKeysPropertyIR(TBL.KEYS)
      => $is_some_<tableEntryPriorityIR>(tableEntryPriorityOptIR)
  -- ...
```

### Binary values in STF files

When providing the value of a key to add a new table entry in STF files, only hexadecimal numbers were supported for ternary keys. Now a binary number may also be provided as the value of ternary keys, as well as lpm keys. For example, consider the table `ipv4_fib_lpm` in `basic_routing-bmv2.p4`:

```p4
control ingress( ... ) {
    ...
    table ipv4_fib_lpm {
        actions = {
            on_miss;
            fib_hit_nexthop;
        }
        key = {
            meta.ingress_metadata.vrf: exact;
            hdr.ipv4.dstAddr         : lpm;
        }
        size = 16384;
    }
    ...
}
```

`testdata/p4testgen/basic_routing-bmv2/basic_routing-bmv2_1.stf` adds an entry, providing the value of the key `hdr.ipv4.dstAddr` in binary:

```
add "main.ig.ipv4_fib_lpm" 1 "hdr.ipv4.dstAddr":0b11111111111111111111111111111111 "meta.ingress_metadata.vrf":0x000 "on_miss"()
```

### Test results

The current simulator test results are the following:

v1model: `[EXCLUDE] 117/1171 (9.99%) [PASS] 933/1171 (79.68%) [FAIL] 121/1171 (10.33%)`
ebpf: `[EXCLUDE] 5/78 (6.41%) [PASS] 70/78 (89.74%) [FAIL] 3/78 (3.85%)`

Overall, 126 tests from the v1model architecture and 1 test from the ebpf architecture pass due to this PR.


spec-concrete/6-dynamic-runtime/6.5.3-object-table.watsup
```diff
@@ -20,6 +23,15 @@ def $tableObject_create_entry_priority(eps) = eps
...
 
+;;; Explicit priority requirement check
+
+dec $needs_priority_tableKeysPropertyIR(tableKeysPropertyIR) : bool
+def $needs_priority_tableKeysPropertyIR(KEY `= `{ tableKeyIR* }) = true
+  -- (if _ `: nameIR_matchKind _ `; = tableKeyIR)*
+  -- if $exists_((nameIR_matchKind = "ternary")*)
+def $needs_priority_tableKeysPropertyIR(tableKeysPropertyIR) = false
+  -- otherwise
+
```
jaehyun1ee, line 40: Isn't it enough to put:

```
dec $needs_priority_tableKeysPropertyIR(tableKeysPropertyIR) : bool
def $needs_priority_tableKeysPropertyIR(KEY `= `{ tableKeyIR* })
    = $exists_((nameIR_matchKind = "ternary")*)
  -- (if _ `: nameIR_matchKind _ `; = tableKeyIR)*
```

spec-concrete/6-dynamic-runtime/6.5.3-object-table.watsup
```diff
@@ -4,7 +4,10 @@
 
 syntax tableEntryPriorityInterface = int?
 
-syntax tableKeyInterface = (nameIR, text)
+syntax tableKeyValueInterface =
+  | `NUM text
+  | text `SLASH nat
```
jaehyun1ee: Nice, can we further refine this with:

```
syntax tableKeyValueInterface =
  | `BIN text
  | `HEX text
  | `DEC text (* I wonder whether this is used *)
  | text `SLASH nat
```

so that we can mark whether a table key is bin/hex/slash from the simulator side and let the spec do pattern-matching on the ADT, instead of having to do prefix check of "0b" and "0x"?

jaehyun1ee (changes requested):
Thanks! Requesting minor changes to the spec.

spec-concrete/6-dynamic-runtime/6.5.3-object-table.watsup
shyukahn: Yes, you're right. Updated in [b6d53bb](https://github.com/kaist-plrg/p4-spectec/pull/102/commits/b6d53bbcfde8b47195df07fe5da8c45ee1058d54).

spec-concrete/6-dynamic-runtime/6.5.3-object-table.watsup
shyukahn: Doing pattern-matching looks much clearer than checking the prefix of a raw string. Updated in [b6d53bb](https://github.com/kaist-plrg/p4-spectec/pull/102/commits/b6d53bbcfde8b47195df07fe5da8c45ee1058d54).
I also added another helper function `text_of_number_tableKeyValueInterface()` to prevent adding almost-duplicate rules for the match kind `exact`.
Decimal key values are rarely used, but there is a test case [key-bmv2.stf](https://github.com/kaist-plrg/p4c/blob/ba98dee3bd8df212217f1ea8a1d61817a2a7a8e4/testdata/p4_16_samples/key-bmv2.stf).

jaehyun1ee (approved):
LGTM! Will merge soon ...

PR #105, opened 2026-01-23
Extern function `resubmit_preserving_field_list`

pacokwon:
## `resubmit_preserving_field_list` implementation
* Update `fieldTypeIR` to include annotationList
* Support field list preservation
* Update pipeline logic for resubmit
* 2 new STF tests, as no other existing tests reach `resubmit_preserving_field_list`

## Correct use of `$` in STF format
Each entry in the expect queue also has a comparison policy. Our previous comparison policy is only used when there is a `$` at the end of the expected packet output. When there is no `$`, the expect packet must be a prefix of the output packet.


p4spec/lib/backend-sim/v1model/pipe.ml
```diff
@@ -445,6 +472,39 @@ struct
...
+        let packet_in = get_packet_in value_sto in
+        let packet_in = Core.Object.PacketIn.reset packet_in in
+        let packet_in = PacketIn packet_in in
+        let value_objectId = wrap_list_v "id" [ wrap_text_v "packet_in" ] in
+        let value_packet_in =
+          packet_in |> extern_to_yojson |> wrap_extern_v "externState"
+        in
+        Spec.Func.update_store_externState value_sto value_objectId
+          value_packet_in
+      in
```
jaehyun1ee, line 497: We should use the most recent store, provided from the `Resubmit` exception. But because the parser block has to proceed again, we should reset the `packet_in` object's state. The job is done here.

jaehyun1ee (approved):
LGTM! Made minor nits myself. PTAL at how I edited the resubmit part in V1Model.

p4spec/lib/backend-sim/v1model/pipe.ml
pacokwon: Sounds good!

PR #107, opened 2026-01-24
Add subtype check expression to EL

jaehyun1ee:
This PR adds `SubE` to EL, which checks whether an expression is a subtype of a meta-type. It uses syntax `exp <: plaintyp`, for example,

```
rule ReturnType_wf:
  B |- typeIR_ret
  -- Type_wf: B |- typeIR_ret
  -- if typeIR_unroll = $unroll_typeIR(typeIR_ret)
  -- if ~(typeIR_unroll = INT)
        /\ ~(typeIR_unroll <: synthesizedTypeIR)
        /\ ~(typeIR_unroll <: objectTypeIR)
```

Elaboration checks that the type of `exp` and `plaintyp` are comparable types. Subtype check for incomparable types, e.g., `"foo" <: bool` is rejected during elaboration. After elaboration, `SubE` in EL becomes `SubE` in IL.

Note that `SubE` already existed in IL/SL. Previously, `SubE` were only generated internally after partial bind analysis. For instance, a premise `-- Lvalue_read: p EC_1 STO_1 |- storageReference : headerStackValue` would internally become:

```
-- Lvalue_read: p EC_1 STO_1 |- storageReference : value
-- if value <: headerStackValue
-- let headerStackValue = (headerStackValue) value
```

Yet, because `SubE` were only internally generated, to express its negation (`exp` is *not* a subtype of `plaintyp`), we had to rely on auxiliary functions:

```
dec $is_headerStackValue(value) : bool
def $is_headerStackValue(headerStackValue) = true
def $is_headerStackValue(value) = false
   -- otherwise

  rule Lvalue_eval/cont:
    p EC_0 STO_0 |- (typedLvalueIR `. nameIR) `# _
                  : EC_1 STO_1 storageReferenceResult
    -- Lvalue_eval:
        p EC_0 STO_0 |- typedLvalueIR
                      : EC_1 STO_1 (`` storageReference)
    -- Lvalue_read:
        p EC_1 STO_1 |- storageReference : value
    -- if ~$is_headerStackValue(value)
    -- if storageReferenceResult
        = `` (storageReference `. nameIR)
```

Because `-- if value <: headerStackValue` and `if ~$is_headerStackValue(value)` are syntactically quite different, structuring failed to pair the two as If-then-Else. Now with `SubE` expressible in EL, the spec can be expressed as:

```
  rule Lvalue_eval/cont:
    p EC_0 STO_0 |- (typedLvalueIR `. nameIR) `# _
                  : EC_1 STO_1 storageReferenceResult
    -- Lvalue_eval:
        p EC_0 STO_0 |- typedLvalueIR
                      : EC_1 STO_1 (`` storageReference)
    -- Lvalue_read:
        p EC_1 STO_1 |- storageReference : value
    -- if ~(value <: headerStackValue)
    -- if storageReferenceResult
        = `` (storageReference `. nameIR)
```

Thus, auxiliary functions are no longer necessary, and structuring finds more condition pairs. Alongside this, `SubE` in both IL and SL interpreters have been made more robust, with a slight performance trade-off.


jaehyun1ee:
@cfbolz for visibility. `SubE` is not entirely new, but I've added some more cases to the IL/SL interpreter for evaluating `SubE`.

PR #109, opened 2026-01-26
Optional paramters annotated by the `@optional` annotation

shyukahn:
This PR allows parameters to be optional when annotated by the `@optional` annotation.

## `@optional` annotations

As per the [P4 specification 6.8.2](https://p4.org/wp-content/uploads/sites/53/p4-spec/docs/p4-16-working-draft.html#sec-optional-parameters), a parameter that is annotated with the `@optional` annotation is optional.

The implementation is mostly done in overload resolution, i.e. `4.2-ir-call-overload.watsup`. When the arity of parameters and arguments don't match, all of the omitted parameters were previously considered to be parameters with default values. Now optional parameters annotated by `@optional` are also considered, allowing such parameters to be omitted.

From the spec, not all parameters can be annotated as optional:
> Optional parameters can only appear for arguments of: packages, parser types, control types, extern functions, extern methods, and extern object constructors.

To distinguish the sites where parameters may be optional, a new parameter context `paramctxt` must be provided to the parameter typing relations:
```
syntax paramctxt = OPT | NOOPT

relation Parameter_ok:
  cursor typingContext paramctxt |- parameter
                                  : typingContext parameterIR `# typeId*

relation ParameterList_ok:
  cursor typingContext paramctxt |- parameterList
                                  : typingContext parameterListIR `# typeId*

relation ConstructorParameterListOpt_ok:
  typingContext paramctxt |- constructorParameterListOpt
                           : typingContext constructorParameterListIR `# typeId*
```

Note that there are several more restrictions on optional parameters from the spec, such as:
* Optional parameters cannot have default values.
* Constructs with both optional parameters and default values can only be called using named arguments.

Additionally, some stronger constraints are added. The following are allowed in P4C (though ambiguous in the spec), but are decided to be disallowed in P4-SpecTec:
* Optional parameters in parser/control declarations.
* Optional parameters in abstract methods.

Optional parameters are allowed in parser/control type declarations, but not in parser/control declarations. For example, [issue3051.p4](https://github.com/kaist-plrg/p4c/blob/ba98dee3bd8df212217f1ea8a1d61817a2a7a8e4/testdata/p4_16_samples/issue3051.p4) is excluded:
```p4
control MyC(@optional inout bool b) {
  apply {
     b = !b;
  }
}
```

Optional parameters in abstract methods are also not allowed, such as in [issue3417.p4](https://github.com/kaist-plrg/p4c/blob/ba98dee3bd8df212217f1ea8a1d61817a2a7a8e4/testdata/p4_16_samples/issue3417.p4), which is excluded:
```p4
extern RegisterAction<T, I, U> {
    ...
    abstract void apply(inout T value, @optional out U rv);
}
```

### Function `$align_parameterListIR()`

`$align_parameterListIR()` returned 4 values in a tuple, and one of them was `argumentIR*`, which is the list of the aligned arguments. However, `$align_parameterListIR()` aligns parameters with respect to the arguments without modifying the order of the arguments. That is, returning the arguments was redundant. Now it returns 3 values in a tuple:

```
;; Before
dec $align_parameterListIR(parameterIR*, argumentIR*, nameIR*)
  : (parameterIR*, argumentIR*, parameterIR*, value*)

;; After
dec $align_parameterListIR(parameterIR*, argumentIR*, nameIR*)
  : (parameterIR*, parameterIR*, value*)
```
Note that the second `parameterIR*` is the list of the default parameters, which doesn't include the optional parameters annotated with `@optional`.

### Variable naming

All of the omitted parameters were previously considered as default parameters, but now each of them are one of a default parameter or a parameter annotated with `@optional`. Considering this, variables that denote optional parameters are renamed if they include `default` in their names. For example, many use cases of `id_default` are renamed to `id_omitted`.

## Test results

3 test cases were removed from `excludes/static/target-specific/annotation.exclude`:
* `gauntlet_optional-bmv2.p4`
* `issue2664-bmv2.p4`
* `named-arg1.p4`

Some test cases with `@optional` are still excluded by the following reasons:
* Non-inferrable type parameters
  - `issue1304.p4`
  - `issue2492.p4`
  - `issue2630.p4`
  - `issue2810.p4`
* `@optional` in abstract methods
  - `issue3417.p4`
  - `issue2273-1.p4`
* `@optional` in parser/control declarations
  - `issue3051.p4`

`@optional` in abstract methods and parser/control declarations is described above.

Consider [issue1304.p4](https://github.com/kaist-plrg/p4c/blob/ba98dee3bd8df212217f1ea8a1d61817a2a7a8e4/testdata/p4_16_samples/issue1304.p4):
```p4
package Switch<H0, M0, H1, M1>(Pipeline<H0,M0> p0, @optional Pipeline<H1,M1> p1);
Pipeline( ... ) p0;
Switch<my_packet, my_metadata, _, _>(p0) main;
```
`p1` is an optional parameter of the package `Switch`. The declaration of `main` omits this, and the type parameters `H1` and `M1` are given as "don't care" types, i.e. omitted and must be inferred. But since `p1` is not given, type inference fails.


jaehyun1ee (approved):
Thanks! I made some refactoring myself:

- Whether optional parameter is allowed (e.g., for an extern function or a parser apply method) was previously checked in `ParameterType_ok` via `paramctxt`, but now this check is moved to `CallableType_wf` and `ConstructorType_wf`.
- `# id_default*` and `# id_omitted*` were used throughout the spec, which I found a bit confusing, so I explicitly disambiguated them as `# id_default* # id_optional*`. The spec got verbose, but is explicit about what kind of ids are used.

Merging, but @shyukahn PTAL and ping if there's something wrong with the refactored spec :)

shyukahn:
Thanks for your review and changes! The refactored spec looks much better to me.

PR #111, opened 2026-01-29
Packet Scheduler for V1Model

pacokwon:
## Adding `archState` to `store`
The `store` type in the spec now contains `archState` - an extern type specifically for storing architecture specific state.
```
;; Architecture-specific state

extern syntax archState
extern dec $empty_archState : archState

;; Map from object identifiers to objects

syntax objectStore = map<objectId, object>

syntax store = archState objectStore

var STO : store

dec $empty_store : store
  hint(prose_in "an empty store")
def $empty_store = archState objectStore
  -- if archState = $empty_archState
  -- if objectStore = $empty_map<objectId, object>
```

The previously named `externState` type, along with its corresponding functions and values, have been renamed to `objectState`.

Each architecture provides an empty `archState` value. EBPF and Placeholder simulators return a unit value for now.

V1Model defines the following architecture-specific state.
```ocaml
module ArchState = struct
  type t = {
    queue : Scheduler.t;                          (* double ended packet queue *)
    mirror_tbl : MirrorTable.t;                   (* session - port mappings *)
    clone_opt : Packet.CloneInfo.t option;        (* pending clone task *)
    resubmit_opt : Packet.ResubmitInfo.t option;  (* pending resubmit task *)
  }
  (* ... more code ... *)
end
```

As `clone_opt` and `resubmit_opt` is local to each block (more specifically, ingress and egress), it is worth considering refactoring to achieve logical separation of these fields from `queue` and `mirror_tbl`, which are global states.

## Multiple output packets per input packet
Clone operations within a pipeline could potentially lead to multiple output packets per input packet. To reflect this behavior, `drive_pipe` now returns `IO.tx list` instead of `IO.tx option`. Accordingly, the EBPF simulator returns `[tx]` for `Some tx` and an empty list for `None`.

## A Monadic Pipeline
As threading `value_ctx` and `value_sto` is a common pattern in SpecTec, the same is seen in OCaml code dealing with relations and functions. However, calling multiple relations/functions in succession makes the code verbose.

The usage of option + state monad is proposed, in an attempt to mitigate the efforts of managing context and store values in OCaml (+ for fun).

```ocaml
(* lib/backend-sim/v1model/pipe.ml *)

type 'a pipe_ctx = (Value.t * Value.t * IO.tx list, 'a) OptionState.t
(*                  context   store     packet outputs              *)
```

This allows the context and store to be accessed only when they are needed. Reducing their need to be bound to a variable when unnecessary.

Here is an example of a function threading the context and store repeatedly:
```ocaml
let drive_pipe_pre_inner (value_ctx : Value.t) (value_sto : Value.t) :
    Value.t * Value.t * bool =
  (* Parser block *)
  let value_ctx, value_sto = drive_p value_ctx value_sto in
  (* Verify block *)
  let value_ctx, value_sto, _value_verify_result =
    drive_vr value_ctx value_sto
  in
  (* Ingress block *)
  let value_ctx, value_sto, _value_ingress_result =
    drive_ig value_ctx value_sto
  in
  (* ... *)
```

Here is the same function using `pipe_ctx`
```ocaml
let drive_pipe_pre_inner : bool pipe_ctx =
  let* () = drive_p in
  let* _value_verify_result = drive_vr in
  let* _value_ingress_result = drive_ig in
  (* ... *)
```

or even shorter:
```ocaml
let drive_pipe_pre_inner : bool pipe_ctx =
  let* _value_ingress_result =
    drive_p
    >> drive_vr
    >> drive_ig
  in
  (* ... *)
```

Currently, `lib/backend-sim/v1model/pipe.ml` is refactored to use the `pipe_ctx` monad where it could benefit the most. There is room for more proactive refactoring.

## Packet Scheduler
To support the V1Model primitive `clone`, a scheduler is added to the V1Model simulator.

A packet is ran in one of three execution phases (it will not be interrupted in the middle of the phase)
1. Parser + Verify
2. Ingress
3. Egress + Update + Deparser

After phases 1 or 2, a packet is scheduled to an *entrypoint*. An *entrypoint* is where the packet will resume processing the next time it is selected by the scheduler. An *entrypoint* is either `Ingress` or `Egress`.

Normally, a packet goes through phase 1. Then it is scheduled to start at `Ingress`. Then it is scheduled to start at `Egress`, where it finishes being processed.

### Resubmit

When a packet is resubmitted in ingress, it is *not* scheduled to start at `Egress`. Instead,
1. Its packet input is reset,
2. Goes through phase 1
3. Scheduled to start at ingress

### Clone
When a packet is cloned in ingress:
1. Its packet input is reset
2. Goes through phase 1
3. Scheduled to start at egress

When a packet is cloned in egress, it is simply scheduled to start at egress again.

## STF
Support for `mirroring_add` is added. `mirroring_add [session] [port]` adds a mapping from a *mirroring session* to an output port. Whenever a cloned packet is assigned to a mirroring session, it is sent to the corresponding output port.

NOTE: A function `val add_mirror_session : Value.t -> int -> int -> Value.t` has been added to the `ARCH` type. However, `mirroring_add` seems to be specific to BMv2, and not to STF (not that STF has a standard 🥺). Since there is no 해당사항 for other architectures, it might be better to find a smarter way for V1Model to run `mirroring_add`.

There are STF files that has not been patched, due to the patcher skipping those that had `mirroring_add` in the past. These files have been patched to use qualified names for tables and unqualified names for actions.


jaehyun1ee (approved):
Thank you for the commits, I made some changes myself alongside minor nits. The main changes are:

### Pulled `archState` out of `store`

I believe that conceptually, `archState` should be set-up after instantiation and right before execution. So I pulled `archState` out of `store`, s.t. instantiation only deals with `map<objectId, object>`, and created `syntax arch` for dynamic semantics that contains both the map and `archState`.

```
;; 7.01.2-dynamic-store.watsup

syntax store = map<objectId, object>

;; 8.01.2-eval-arch.watsup

extern syntax archState
extern dec $init_archState : archState

syntax arch =
  { STATE archState,
    STORE store }
```

### Moved `option_state` monad from `utils` to `backend_sim`

The monadic refactoring indeed seems nice! I think the best part is that function signatures are no longer so lengthy.

Though I don't see any immediate use case of this monad outside `backend_sim`. So I moved it from `utils` to `backend_sim`, and partly specialized the type as:

```ocaml
(* backend_sim/state.ml *)

(* The state consists of:
    - current evaluation context
    - current arch
    - list of transmissions to perform *)

type s = Value.t * Value.t * IO.tx list

(* State monad with failure *)

type 'a state = s -> 'a option * s
```

And also added one more helper, which used to be `with_pipe_ctx` before.

```ocaml
(* Apply: apply a function to the current context and arch, updating them *)

let apply (f : Value.t -> Value.t -> Value.t * Value.t * 'a) : 'a state =
  let* value_ctx, value_arch, txs = get in
  let value_ctx, value_arch, result = f value_ctx value_arch in
  let+ () = put (value_ctx, value_arch, txs) in
  result
```

Merging now, but PTAL @pacokwon.

PR #112, opened 2026-01-30
PSA architecture support

shyukahn:
This PR adds support for the PSA, following the [PSA specification](https://p4.org/wp-content/uploads/sites/53/p4-spec/docs/psa-working-draft-html-version.html).

## Portable Switch Architecture (PSA)

The PSA model has six programmable P4 blocks and two fixed-function blocks. The ingress parser block, ingress block, and ingress deparser block form the ingress pipeline, and the egress parser block, egress block, and egress deparser block form the egress pipeline. The Packet buffer and Replication Engine (PRE) and the Buffer Queuing Engine (BQE) are target dependent functional blocks. If a packet is not dropped/cloned/resubmitted/recirculated, it goes through the following order:

1. Ingress pipeline
2. PRE
3. Egress pipeline
4. BQE

Ingress/egress metadata can be modified during the pipeline evaluation to specify the packet path. Details on packet paths are found in [3. Packet paths](https://p4.org/wp-content/uploads/sites/53/p4-spec/docs/psa-working-draft-html-version.html#packet-paths).

## Implementation

Pipeline initialization and the invocations of P4 blocks are written in the following files:
* `spec-concrete/9-arch/9.C.1-psa.watsup`: PSA initialization, with some common helper functions.
* `spec-concrete/9-arch/9.C.2-psa-ingress-pipeline.watsup`: Ingress pipeline initialization/invocation.
* `spec-concrete/9-arch/9.C.3-psa-egress-pipeline.watsup`: Egress pipeline initialization/invocation.

The pipeline simulator is implemented in `p4spec/lib/backend-sim/psa/pipe.ml`.

The following extern blocks are implemented in `p4spec/lib/backend-sim/psa/object.ml`:
* `Counter`
* `Meter`
* `Register`
* `Hash`
* `InternetChecksum`

## Packet scheduling

After a packet passes the ingress or egress pipeline, the scheduler schedules the packet by reading the metadata fields.

### Normal unicast

After the ingress pipeline, a packet is sent to the egress pipeline. This is the "normal" path of packets.

### Normal multicast

After the ingress pipeline, a packet is sent to a multicast group if a multicast group id is specified in its metadata. Multicast groups are configured in the control plane. STF commands `mc_group_create`, `mc_node_create`, and `mc_node_associate` configure multicast groups, in the same way as the v1model architecture. From the multicast group, 0 or more packets are sent to the egress pipeline.

### Resubmit

After the ingress pipeline, a packet may be resubmitted to the ingress pipeline again. The content of a resubmitted packet is same as the original packet _before_ it had been transferred to the ingress pipeline, i.e. modifications by the ingress pipeline are ignored.

### Clone I2E

After the ingress pipeline, a packet may be cloned with a clone session id in its metadata. Clone sessions are configured in the control plane, and their structure are similar to multicast groups. A clone session produces 0 or more cloned packets and sends them to the egress pipeline. The STF command `mirroring_add_mc` associates a clone session id to a multicast group id.

Note that the main difference between multicast and clone i2e is the content of a cloned packet. Similar to resubmitted packets, the content of a packet that is cloned after the ingress pipeline is same as the original packet _before_ it had been transferred to the ingress pipeline, i.e. modifications by the ingress pipeline are ignored. On the other hand, the content of multicasted packets are modified by the ingress pipeline.

### Clone E2E

After the egress pipeline, a packet may be cloned and be sent to the egress pipeline again. The identical clone session configurations are used with clone i2e. The content of a cloned packet is same as the resulting packet after processing the egress pipeline.

### Recirculate

After the egress pipeline, a packet may be recirculated to the ingress pipeline. The content of a recirculated packet is same as the original packet after processing the egress pipeline.

## STF

The following STF commands are now supported:
* `mirroring_add_mc <clone_session_id> <multicast_group_id>`: Associates the configuration of the given clone session with the multicast group.
* `register_read <register_name> <index>`: Reads the value of a register at the given index. For now, the command does read the register but does not return it in some way.
* `register_write <register_name> <index> <value>`: Writes the given value to a register at the given index.
* `register_reset <register_name>`: Resets all values of a register to default.

## Commands

To run the specification with the PSA:

```bash
$ ./p4spectec sim spec-concrete/*/*.watsup -arch psa -i p4c/p4include -p [FILENAME].p4 -stf [STF FILENAME].stf -sl
```

`make test-sim-psa-p4c-sl` runs the test for the PSA.

## Test results

In `p4c/testdata/p4_16_samples`, there exists 26 pairs of P4 files and STF files that target the PSA. Among them, 2 tests are excluded:
* `psa-bool-ternary-const-entry-bmv2.p4`: Excluded before type-check. It is already included in `excludes/static/bug/ternary-match-on-bool.exclude`.
* `psa-meter7-bmv2.p4`: Excluded due to unsupported STF commands (`meter_set_rates` / `meter_get_rates` / `meter_array_set_rates`) in `psa-meter7-bmv2.stf`. It is added to `excludes/dynamic/stf/command-not-implemented.exclude`.

Current test results: `[EXCLUDE] 2/26 (7.69%) [PASS] 24/26 (92.31%) [FAIL] 0/26 (0.00%)`

## Miscellaneous

Previously, a parser value set was initialized with `n` elements, each set to the default value, where `n` is the size. This behavior has changed so that value sets are now initially empty. Consequently, `value_set_ebpf.p4` now passes the eBPF architecture simulator test.


jaehyun1ee (approved):
LGTM! Thanks for the addition!

PR #114, opened 2026-02-10
Multicast and more V1Model constructs

pacokwon:
### Name annotation
Control plane operations can now use aliases (through the @name annotation) for the following:
* Actions
* Action parameters
* Tables

### Unroll newtype during table entry addition
There are a few programs that had keys whose types are newtypes. While casting an underlying type to a newtype is conceptually odd, it is performed (only) when control plane operations assign values to them.

### Multicast
Control plane operations `mc_mgrp_create`, `mc_node_create` and `mc_node_associate` has been implemented.

`mc_mgrp_create` creates a "multicast group". `mc_node_create` creates a "node". And `mc_node_associate` puts a node under a group. When a packet is assigned to a multicast group, it is replicated and the copies are assigned their own output ports. The number of replicated packets and their ports are defined by the nodes under the multicast group.

These commands, along with `mirroring_add` are BMv2 specific. I have considered making a single entrypoint for architecture-specific control plane operations so that each architecture implementation can handle them accordingly. That isn't currently implemented, but let me know your thoughts on that.

### Recirculate
The `recirculate_preserving_field_list` extern function is implemented. When it is called, instead of being transmitted, the output packet from the deparser block is fed back into the pipeline, with certain metadata fields preserved.

### Packet matching policy
Conceptually, the V1Model architecture has a queue for each output port. This is worth noting because it is only within each queue that the order of the output packets matter.

For example, a program that passes this packet input should not pass when the two `expect` lines are swapped.
```
packet 0 00
expect 1 00 $
expect 1 11 $
```

However, a program that passes this packet input should pass when the two `expect` lines are swapped.
```
packet 0 00
expect 1 00 $
expect 4 11 $
```

The matching policy has been updated to emulate this behavior.

### Patches
The following STF files have been patched.
* `testdata/p4testgen/issue298-bmv2/inline-bmv2_1.stf`: rename `$valid$` -> `isValid<>()`
* `testdata/p4testgen/issue298-bmv2/issue298-bmv2_.stf`: rename `intrinsic_metadata` -> `local_metadata`
* `patches/v1model/ipv6-switch-ml-bmv2.stf`: use qualified names for tables, unqualified names for actions
* `patches/v1model/v1model-special-ops-bmv2.stf`: use qualified names for tables, unqualified names for actions


jaehyun1ee (approved):
LGTM! Made minor nits myself.

NOTE: You can execute `make test-all-det` to run IL/SL tests with the `-det` flag to find non-determinism bugs.

PR #115, opened 2026-02-15
IL/SL updates: special treatment of `otherwise`, structured `LetI`/`RuleI`, validation for non-determinism, improved caching, and test refactorings

jaehyun1ee:
This is a major update on IL/SL and build tests. Changes are:

1. `otherwise` are treated as special kinds, different from ordinary premises in IL and instructions in SL
2. `LetI` and `RuleI` are now structured in SL, i.e., a match failure during evaluation of `LetI` and `RuleI` can now be expressed at SL-level
3. IL/SL interpreters now have a `-det` mode, which checks whether a non-deterministic match occurs during evaluation
4. Improve caching (*Added 2/19*)
5. Build tests are modularized as part of refactoring (*Added 2/19*)
6. Build tests do not print source location or error logs to stdout to debloat diffs (*Added 2/19*)
7. Three kinds of build tests are offered: `make test-fast`, `make test-all`, and `make test-all-det` (*Added 2/19*)

### 1. Special treatment of `otherwise`

```
def $subst_typeIR(`{ eps }, typeIR) = typeIR
def $subst_typeIR(theta, typeIR) = $subst_typeIR'(theta, typeIR)
  -- otherwise
```

A rule/clause with an `-- otherwise` premise is special, since it acts as a final rule that is applied when *all other* rules/clauses fail. For example, for `$subst_typeIR`, the second clause applies when the first clause fails. Yet, it was previously unchecked that: (i) `-- otherwise` should not be accompanied with other side-conditions or side-effects, and (ii) `OtherwiseI` must be the last instruction in SL.

For (i), consider the case:

```
def $subst_typeIR(`{ eps }, typeIR) = typeIR
def $subst_typeIR(theta, typeIR) = $subst_typeIR'(theta, typeIR)
  -- if ... some condition ...
  -- if $partial_function(...)
  -- otherwise
```

An `-- otherwise` should *not* be used alongside other side-condition or a side-effecting premise (in the example, the premise may fail due to its partiality). This wasn't checked, and now the elaboration pass validates this.

For (ii), consider the case:

```
def $subst_typeIR(theta, typeIR) = $subst_typeIR'(theta, typeIR)
  -- otherwise
def $subst_typeIR(`{ eps }, typeIR) = typeIR
``` 

This is also valid, as there is no explicit order between inference rules and clauses. (We *assume* that all rules and clauses are mutually exclusive, thus they are always matched to a single rule/clause or none.) Previously, structuring neglected such a case, and would generate an instruction sequence starting with `OtherwiseI`, which is wrong. Now, `-- otherwise` is treated as a special kind:

```ocaml
(* lang/il/ast.ml *)
type def =
  | RelD of id * nottyp * Hints.Input.t * rulegroup list * elsegroup option * hint list
  | FuncDecD of id * tparam list * param list * typ * clause list * elseclause option * hint list
  | ...

(* lang/sl/ast.ml *)
type rel = id * rel_signature * exp list * block * elseblock option * hint list
type definedfunc = id * tparam list * param list * typ * block * elseblock option * hint list
```

### 2. Structured `LetI` and `RuleI`

Structuring produces SL from IL, primarily to couple mutually exclusive premises as an `if-else` structure. At a high level, for each rule, all premises other that `-- if` were linearized, then concatenated into a single sequence of instructions.

Yet, this conversion failed to preserve the semantics in IL, where in IL, match failure of a relation/function is also part of the backtracking semantics. For example,

```
;; assm. relation Rel is given, and is partial

relation RelCaller:
   |- nat : nat
   hint(input %0)

rule RelCaller/if-Rel-succeeds:
   |- n_a : n_res
   -- Rel: |- n_a : n_res ;; (A)

rule RelCaller/if-Rel-fails:
  |- n_a : n_res
  -- if n_res = 0
```

In IL, if premise (A) fails due to a match failure, it would backtrack and then try rule `if-Rel-fails`. Yet, the linearization during IL-to-SL previously led to the following SL spec:

```
relation RelCaller(n_a)

1. Let n_res = Rel(n_a)
8. Result in n_res
9. Let n_res = 0
10. Result in n_res 
```

This fails to represent the case when `Rel(n_a)` fails. Thus, in sequential evaluation, if `Rel(n_a)` fails, the entire `RelCaller(n_a)` would result in an error. Said otherwise, the previous IL-to-SL pass *assumed* that all relation and function invocations *must* match.

To faithfully represent the partiality of relations and functions, now in SL, all `LetI` and `RuleI` have a nested block, evaluated when the assignment succeeds. At OCaml type level,

```ocaml
(* before *)
and instr' =
  (* Binding instructions *)
  | LetI of exp * exp * iterinstr list
  | RuleI of id * notexp * Hints.Input.t * iterinstr list
  | ...

(* after *)
and instr' =
  (* Binding instructions *)
  | LetI of exp * exp * iterinstr list * block
  | RuleI of id * notexp * Hints.Input.t * iterinstr list * block
  | ...
```

So `RelCaller` would now be:

```
relation RelCaller(n_a)

1. Let n_res = Rel(n_a)
  1. Result in n_res
2. Let n_res = 0
  1. Result in n_res 
```

This also comes at a cost that there are way more nested levels in SL. For now, when generating PL from SL, all `LetI` and `RuleI` nestings are linearized. So, PL still relies on the assumption that relation and function calls must success. There may be two ways to address this: (i) redesign PL, similar to the redesign of SL, or (ii) validate that the assumption holds.

### 3. Validation for non-determinism

Relations and functions are defined via separate rules and clauses in P4-SpecTec. And for efficient algorithmic evaluation w/o having to search, it assumes that for a given relation/function, *at most one* rule or clause matches. This assumption is difficult to check statically, as it requires static reasoning over general boolean-typed meta-expressions to see if they are mutually exclusive.

To check this assumption dynamically, both IL and SL interpreters can now be run with `-det` mode, where it tries matching *all* rules and clauses in a relation or a function. If it detects more than one match, it is an error. This is in contrast with the usual mode of execution, where matching is done sequentially and it succeeds greedily at the moment it first succeeds matching.

`-det` mode found several (yet trivial) bugs in the current spec, and all of them have been fixed. ~Note that `-det` can be exponential, thus IL interpreter hangs on P4 dynamic tests. One way to address this might be using caches.~

Because this is quite a big change to the framework, will keep this open until other PRs are merged.

### 4. Improve caching (*Added 2/19*)

Previously, caches were used within IL/SL interpreters, but in a very limited fashion. Only some carefully-selected functions and relations were cached (via hardcoded names), because (i) hashing values were expensive, and (ii) caching any kind of relation/function led to correctness issues in some cases.

To make hashing less expensive, this PR implements incremental hashing, where value hashes are annotated alongside their respective nodes:

```ocaml
and vid = int
and vnote = { vid : vid; typ : typ'; vhash : int } [@@deriving yojson]
and value = (value', vnote) note [@@deriving yojson]
```

And to safeguard caching from introducing unexpected behavior, this PR disallows caching of:

* relations/functions with side-effects

```ocaml
          let builtin_ctr_before = !Builtin.Fresh.ctr in
          let values_output = invoke_defined_rel' () in
          let builtin_ctr_after = !Builtin.Fresh.ctr in
          (* Cache if the relation does not create a side-effect *)
          if builtin_ctr_before = builtin_ctr_after then
            Cache.Cache.add !rel_cache (id.it, values_input) values_output;
```

* high-order functions
* anonymous functions that are passed as arguments

### 5. Build tests are modularized (*Added 2/19*)

There are now several options for running a test, e.g., choosing between IL/SL and choosing to turn on `-det`. This creates quite a lot of combinations, which is difficult to track within a single `dune` file. So, tests are modularized under subdirectories in `p4spec/test`.

### 6. Build tests do not print source location or error logs to stdout to debloat diffs (*Added 2/19*)

`speclang` tests omit source span and failed `run` or `sim` omit error messages, to prevent minor spec edits turning out into huge diffs overall.

### 7. Three kinds of build tests are offered: `make test-fast`, `make test-all`, and `make test-all-det` (*Added 2/19*)

* `make test-fast` runs all `speclang`, `parse`, `run`, and `sim` tests for only SL without `-det`.
* `make test-all` runs all `speclang`, `parse`, `run`, and `sim` tests for both IL/SL without `-det`.
* `make test-all-det` runs all `speclang`, `parse`, `run`, and `sim` tests for both IL/SL with `-det`.


jaehyun1ee:
A brief summary of perf improvement on running `Program_inst` build tests on 1283 P4 programs (in seconds, and max time per test in parentheses):

| Version/Mode | IL interp | SL interp |
| ------- | -------- | --------- |
| Prev (no `-det`) | 590 (39.83) | 211 (6.91) |
| New (no `-det`) | 188 (8.50) | 120 (3.37) |
| New (`-det`) | 263 (12.22) | 130 (4.11) |

jaehyun1ee:
Now merging. Mentioning @KunJeong @shyukahn @pacokwon @cfbolz @3tty0n for visibility.

PR #118, opened 2026-02-23
Add missing tests for V1Model

pacokwon:
This PR adds 61 new tests from P4Testgen and includes 13 existing tests that have previously been omitted in the testing process. Overall, 73 tests run correctly against P4-SpecTec. 1 test is excluded due to it expecting a specific behavior upon index access out of bounds, which according to the spec is undefined behavior.

# More P4Testgen generated tests
Previously, we added P4Testgen generated tests for P4 programs in the p4c repository that lacked a corresponding STF file.

In that process, I omitted some programs that target V1Model. These programs had `#include "v1model.p4"` when I only included programs with a `#include <v1model.p4>`. As a result, 61 additional STF files have been generated and added to our test suite. Like other tests from P4Testgen, they are patched to use fully qualified names for tables and unqualifed names for actions.

# Bug in the P4-STF match policy
There was a bug in the test matching policy. It did not match p4 programs that were in a subdirectory, such as `parser-inline/parser-inline-test1.p4`. This bug has been fixed, and now 13 more pairs in the p4c test suite are searched and tested correctly.


jaehyun1ee (approved):
LGTM! Thanks for spotting the bug.

PR #122, opened 2026-03-09
Support adding table entries in STF files with keys whose match kind is `optional`; various fixes

shyukahn:
This PR adds support of adding table entries in STF files with keys whose match kind is `optional`, defined in the v1model architecture. Additionally, various issues are fixed in the mechanized spec, regarding type casting, serializable enum types, etc.

## Match kind `optional`

The table match kind `optional` is defined by the v1model architecture. It states that a key value is matched by either an `exact` match or a wildcard that matches any value. Adding an entry to a table consisting such a key through the control plane (i.e. STF files) is now supported.

### Type casting in the control plane

When adding table entries, the control plane may provide numerical values to a P4 program to specify key values or action arguments. P4-SpecTec first treats such values in STF files as arbitrary-precision integers, then casts them to their desired types. However, casting failed if an `int` type is not allowed to be implicitly cast to the type. To resolve this issue, a new meta-function `$cast_control_plane` inserts explicit type casting operators to the values provided from STF files when necessary. When it attempts to cast `typedExpressionIR` to `typeIR` and `$cast_unary(typedExpressionIR, typeIR)` fails, it tries to insert more type casting operators when `typeIR` is one of the following:

* A serializable enum type: The expression is casted to the underlying type first before casting it to the enum type.
* A new type introduced by `type`: The expression is casted to the base type first before casting it to the new type.
* An aliased type by `typedef`: The expression is casted to the base type first before casting it to the aliased type. Note that this is implemented to handle cases where the base type is one of the types above; type aliases are also unrolled during implicit type casting.
* A boolean type: The expression is casted to a `bool` type if its type is `int`.

### Type casting and serializable enum types

When a table key's type is a serializable enum, it was casted to its underlying type. This is fixed to retain its original type, and one P4 test file is excluded due to this change.

* `table-entries-ser-enum-bmv2.p4` tries to implicitly cast an arbitrary-precision integer to a serializable enum whose underlying type is a fixed-width integer.
  ```p4
  table t_ternary {
    const entries = {
      (MyEnum1B.MBR1, _)                           : a_with_control_params(1);
      (MyEnum1B.MBR2, MyEnum2B.MBR2 &&& 0xff00)    : a_with_control_params(2);
      (MyEnum1B.MBR2, _)                           : a_with_control_params(3);
    }
  }
  ```

Additionally, some issues regarding type casting from/to serializable enums were fixed:

* Serializable enums should remain the same when casting to its own type.
  ```
  def $cast_enum(ENUM typeId `< _ > `{ _ }, typeId `. nameIR `. value) = typeId `. nameIR `. value
  ```
* Serializable enums should become a singletone set when casting to a set type.
  ```
  def $cast_enum(SET `< typeIR >, enumValue) = SET `{ $cast_op(typeIR, enumValue) }
  ```
* When casting a value whose type is the underlying type of a serializable enum and the value corresponds to the value of an enum field, it should result in the enum value.
  ```
  def $cast_to_enum'(typeId, typeIR, valueFieldIR_h :: valueFieldIR_t*, value_target)
    = typeId `. nameIR `. value  ;; was previously implemented as value_target
    -- if nameIR `= value `; = valueFieldIR_h
    -- if $bin_eq(value, value_target)
  ```

### Miscellaneous

* `$alias_to_original_lookup_map` doesn't fail when an action parameter has an annotation other than `@name`.
* The OCaml implementation of the built-in meta-function `$text_to_int` converts the given string to `Bigint.t` directly using `Bigint.of_string`. It previously converted strings to an OCaml `int` first, which caused runtime errors if the string represents an integer value exceeding the range of `int`.
* `tableKeyIR` is refactored to consist of the name of the key visible to the control plane.
  ```
  ;; before
  syntax tableKeyIR = typedExpressionIR `: nameIR annotationList `;
  ;; after
  syntax tableKeyIR = typedExpressionIR `# nameIR `: nameIR annotationList `;
  ```
  The first `nameIR` stores the key's name, which is either given by the `@name` annotation or is the key expression itself. `testdata/p4testgen/inline-bmv2/inline-bmv2_1.stf` is fixed for this change.

## Test results

One P4 test file `table-entries-ser-enum-bmv2.p4` is excluded in the static test, also excluding 4 STF tests in the dynamic test. 10 STF tests that run against `dash/dash-pipeline-v1model-bmv2.p4` now succeed.

* `run-pos`: `Running interpreter test (Program_inst): [EXCLUDE] 83/1283 (6.47%) [PASS] 1200/1283 (93.53%) [FAIL] 0/1283 (0.00%)`
* `sim-v1model-p4testgen`: `Running simulation test (v1model): [EXCLUDE] 144/1942 (7.42%) [PASS] 1748/1942 (90.01%) [FAIL] 50/1942 (2.57%)`


jaehyun1ee (approved):
LGTM! Thanks for the edits.

PR #123, opened 2026-03-11
Add P4Testgen generated tests for more programs

pacokwon:
This PR adds P4Testgen generated tests for P4 programs in the P4C repository that **already has a corresponding STF file**. Up to 10 tests are produced for each program, and the STF files are patched like other tests that made it into our current test suite.

Updates to the spec has also been made in the process. See commit log for details 🙏


jaehyun1ee (approved):
LGTM! Thanks a lot :)

PR #124, opened 2026-03-11
fix: treat directionless action parameters as DYN not CTK

favxlaw:
## Summary

Directionless parameters of actions were being marked as CTK (compile-time known) 
regardless of context. Per P4-16 spec §18.1, directionless parameters of actions 
are supplied by the control plane at runtime and should not require CTK. Only 
directionless parameters of non-action callables (controls, parsers, functions) 
require CTK.

## Fix

In `$add_parameter_t` (`5.02.1-typing-context.watsup`), check `TC.LOCAL.KIND` 
before assigning CTK/DYN:
- If inside an ACTION → mark directionless parameter as DYN
- Otherwise → keep CTK

This pattern already exists at line 462 of the same file.

## Tests

- `typing-positive/action-directionless-dyn.p4` : action directionless param 
  accepts runtime value (the bug case)
- `typing-negative/control-directionless-ctk-required.p4` :  control directionless 
  param correctly rejects runtime value
- `typing-positive/action-directionless-ctk-also-ok.p4` : action directionless 
  param still accepts compile-time constants (DYN accepts both)

## PNA Test Failures

During investigation  observed that all 31 PNA test programs in 
`p4c/testdata/p4_16_samples/pna-*.p4` fail. After thorough investigation I
confirmed these failures are pre-existing and unrelated to this fix , a minimal 
PNA program with no externs or special features fails identically on the `concrete` 
branch without this fix, with `ConstructorType_ok` failing on `PNA_NIC`. This is 
because PNA has no architecture file in `spec-concrete/9-arch/` unlike V1Model, 
eBPF, and PSA.

Related: p4lang/p4c#5405

### DIFF
--- a/spec-concrete/5-typing/5.02.1-typing-context.watsup
+++ b/spec-concrete/5-typing/5.02.1-typing-context.watsup
@@ -174,6 +174,11 @@ dec $add_parameter_t(cursor, typingContext, parameterIR) : typingContext
   hint(prose_in %2 "added to the" %0 "layer of" %1)
 
 def $add_parameter_t(cursor, TC, _ `EMPTY typeIR id _) = TC'
+  -- if ACTION = TC.LOCAL.KIND
+  -- if varTypeIR = `EMPTY typeIR DYN eps
+  -- if TC' = $add_var_t(cursor, TC, id, varTypeIR)
+def $add_parameter_t(cursor, TC, _ `EMPTY typeIR id _) = TC'
+  -- if ACTION =/= TC.LOCAL.KIND
   -- if varTypeIR = `EMPTY typeIR CTK eps
   -- if TC' = $add_var_t(cursor, TC, id, varTypeIR)
 def $add_parameter_t(cursor, TC, _ direction typeIR id _) = TC'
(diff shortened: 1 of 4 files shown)


jaehyun1ee:
Thank you for making the PR :)

I see that PNA tests are failing, probably due to how they deal with directionless action parameters within their body. Yet, they are unrelated with `spec-concrete/9-arch`, because `9-arch` deals with the *dynamic* semantics of P4. The issue is more about fixing the *static* semantics specification. So, regardless of whether `pna` is in `9-arch`, it should be at least type checked.

I appreciate the changes, but will keep open until the intended semantics of directionless action parameters are resolved either in `p4-spec` or `p4c` (i.e., to confirm that the PNA tests should be amended, or relaxing the specification).

favxlaw:
> Thank you for making the PR :)
> 
> I see that PNA tests are failing, probably due to how they deal with directionless action parameters within their body. Yet, they are unrelated with `spec-concrete/9-arch`, because `9-arch` deals with the _dynamic_ semantics of P4. The issue is more about fixing the _static_ semantics specification. So, regardless of whether `pna` is in `9-arch`, it should be at least type checked.
> 
> I appreciate the changes, but will keep open until the intended semantics of directionless action parameters are resolved either in `p4-spec` or `p4c` (i.e., to confirm that the PNA tests should be amended, or relaxing the specification).

Thank you @jaehyun1ee  for the clarification! I misidentified the root cause. I'll look into how PNA programs use directionless action parameters in their bodies to better understand what resolution is needed upstream. Looking forward to seeing how p4-spec or p4c resolves this.

PR #125, opened 2026-03-12
Patch STF statements at runtime

pacokwon:
This PR replaces the P4Testgen generated portion of our test suite. Tests are no longer processed before being included in the test suite, rather, a transformation on the STF AST is applied by each architecture at runtime before each statement is executed.

`issue298-bmv2_{1,2,4,5,7,8,10}.stf` and `match-on-exprs2-bmv2_{1,2,3,4}.stf` needed manual patching. The patched tests are located in `patches/v1model`.


jaehyun1ee (approved):
Thanks a lot!

PR #128, opened 2026-03-13
Add table entries with qualified action names

pacokwon:
This PR adds supports for adding table entries with qualified action names from the control plane.

Every action entry in a table now stores a control plane name. Upon table entry installation, the control plane will compare a given name to each control plane names within the action list.

Also, `run` tests do not run tests in `testdata/custom` anymore.


jaehyun1ee (approved):
LGTM! Thanks :)

PR #129, opened 2026-03-16
Remove PSA patches; find table/actions by an unqualified name if qualified name fails

shyukahn:
## Remove PSA patches

STF files of the PSA occur on runtime as V1Model and eBPF. All patched STF files located in `patches/psa` are removed:

* `patches/psa/psa-example-register2-bmv2.stf`
* `patches/psa/psa-register-complex-bmv2.stf`
* `patches/psa/psa-register-read-write-bmv2.stf`

## Find table/actions by an unqualified name if qualified name fails

When a table/action name is given in an STF file, it may correspond to an unqualified name though it is formatted as a qualified name, e.g. `c.t` instead of `main.ig.c.t`. To handle such cases, if searching fails using the qualified name, the simulator retries searching using the unqualified name. That is, for example, the table is searched by `t` if `c.t` fails.

One patched STF file is removed:

* `patches/v1model/key-bmv2.stf`

## Use table keys in STF files in order

Some STF files provided wrong names for table keys when adding a table entry in an STF file. To handle such cases, if the key names do not match, the key values are assumed to be given in the same order as the keys are declared in the P4 file. This generalization lets us remove 11 patched STF files:

* `patches/v1model/ternary2-bmv2.stf`
* `patches/v1model/issue298-bmv2/issue298-bmv2_{1, 2, 4, 5, 7, 8, 10}.stf`
* `patches/v1model/match-on-exprs2-bmv2/match-on-exprs2-bmv2_{2, 3}.stf`
* `patches/ebpf/key-issue-1020_ebpf/key-issue-1020_ebpf_2.stf`


jaehyun1ee (approved):
Thanks a lot!

PR #131, opened 2026-03-17
Exit codes for `run` and `sim`

pacokwon:
`./p4spectec run` exits with 42 if well-typed, 6 if ill-typed, 1 for anything else
`./p4spectec sim` exits with 42 if STF test passed, 6 on failure to match packets, 1 for anything else


pacokwon:
Force pushed after rebase :)

PR #132, opened 2026-03-17
Remove non-architecture-specific behavior from architecture implementations and STF-specific behavior from the mechanized spec

shyukahn:
This PR consists of 2 refactorings that improve the codebase.

## Non-architecture-specific behavior

The simulators of the P4 architectures supported by P4-SpecTec (V1Model, eBPF, PSA) had their own initialization processes and table interfaces. But their implementations did not differ from each other, making the code redundant. 3 functions from `module ARCH` in `p4spec/lib/runtime/sim/simulator.ml` are removed.

* `init` is removed. Its previous body is now included in the initialization of the driver.
* `table_add_entry` and `table_add_default_action` are removed and replaced by `Table.add_entry` and `Table.add_default_action`, respectively, implemented in a new file `p4spec/lib/backend-sim/table.ml`.

## STF-specific behavior in the mechanized spec

The `add` command in STF adds an entry to a table, specifying the name and field of each table key. If the given key names do not match with the actual names defined in the table, the names are ignored and it is assumed that the fields are given in order, matching the values with the keys in the table sequentially. This behavior was illustrated in `spec-concrete/9.1-table-interface.watsup` of the mechanized spec.

Since such STF-specific behavior is not appropriate to be part of the specification, the implementation is moved to the simulator backend, written in OCaml.


jaehyun1ee (approved):
LGTM! Thanks

PR #135, opened 2026-03-23
Modularize type checking of tables

### LINKED ISSUE #130: Modularize entry priority checks when type checking tables
The mechanized specification defines a `tableContext` to store the information required for type-checking tables.

```
syntax tableContext =
  { KEYS matchKey*,
    ACTIONS matchAction*,
    PRIORITIES matchPriority,
    ENTRIES matchEntries,
    MODE matchMode }
```

Among the fields, `PRIORITIES` and `MODE` are primarily used to compute the priority of table entries.

```
syntax matchPriority =
  { VALUES nat*,
    INIT bool,
    DELTA nat,
    LARGEST_WINS bool }

syntax matchMode =
  ;; does not require explicit priority and is not lpm
  | NOPRI
  ;; does not require explicit priority and is lpm, holding the max prefix
  | NOPRILPM nat
  ;; requires explicit priority and is not lpm
  | PRI
  ;; requires explicit priority and is lpm
  | PRILPM
```

During table type-checking, a tableContext (denoted as `TBLC`) is initialized, and priority computation proceeds in two steps:

1. **Mode Update**: `TBLC.MODE` is updated while type-checking the table keys. Starting from `NOPRI`, the mode is updated via state transitions based on each key’s match kind.
2. **Priority Calculation**: If required, the priority of each table entry is computed during type-checking. `TBLC.MODE` determines whether a priority value is necessary and, if so, how to it should be calculated. (See [Section 14.2.1.4.1. Entry priorities](https://p4.org/wp-content/uploads/sites/53/p4-spec/docs/p4-16-working-draft.html#sec-entry-priorities)).

While this implementation is functional, the semantics of entry priorities may be difficult for a reader of the mechanized specification to understand. Separating the priority computation logic from the type-checking process would improve the specification's readability.

shyukahn:
This PR refactors the type checking rules of tables by modularizing some processes.

## Remove table match mode `PRILPM`

Within the table context, the match mode was set to one of `NOPRI`, `NOPRILPM`, `PRI`, or `PRILPM`. `PRILPM` states that an explicit priority for each entry is required and there exists a table key whose match kind is `lpm`. But since there was no need to distinguish `PRI` and `PRILPM` when type checking a table, `PRILPM` is removed.

## Update the table match mode after type checking table keys

Previously, the table match mode was updated each time a table key was type checked, referring to the key's match kind each time. The meta-function `$update_mode_tbl` was called within `rule TableKey_ok`.

`$update_mode_tbl` is refactored to be called only once for each table within `rule TableProperty_ok/key`. Given a `tableKeyIR*`, it updates the `tableContext`.

## Set the entry priorities after type checking table entries

When type checking each table entry, its priority was determined simultaneously. To do so, the table context needed to include `PRIORITIES.VALUES` and `PRIORITIES.INIT`. The relation `TableEntry_priority_ok` referred to these values, inserting a new priority value when needed. Several modifications were made to refactor this behavior.

* `TableEntry_priority_ok` does not check whether a priority value is needed to the current table entry.
* `VALUES` and `INIT` are removed from `matchPriority`.
* After type checking the entries of a table, the resulting `tableEntryIR*` is passed to a new meta-function `$set_priorities_of_tableEntryListIR`, where it validates the priorities of each entry and inserts a priority value when needed. It is called once in `rule TableProperty_ok/entries`.
* The signature of the relations `TableEntry_priority_ok` and `TableEntry_ok` are modified.
  ```
  ;; before
  relation TableEntry_priority_ok:
    typingContext tableContext tableEntryState |- tableEntryPriority?
                                                : tableContext tableEntryPriorityOptIR
  relation TableEntry_ok:
    typingContext tableContext |- tableEntry : tableContext tableEntryIR

  ;; after
  relation TableEntry_priority_ok:
    typingContext |- tableEntryPriority? : tableEntryPriorityOptIR

  relation TableEntry_ok:
    typingContext tableContext |- tableEntry : tableEntryState tableEntryIR
  ```
* Since `TableEntry_ok` no longer updates the `tableContext`, no folding is needed in type checking a list of entries. The relation `TableEntries_ok` is removed.


jaehyun1ee (approved):
LGTM, thanks!

PR #139, opened 2026-03-24
Support for slice access via `+:`

### LINKED ISSUE #39: Support `+:` slices
This PR adds `+:` slices, where `[i +: n]` represents slicing from `i` with length `n`: https://github.com/p4lang/p4-spec/pull/1316. The standard semantics of it should be expressed in the mechanized spec.

> A slice may also be specified as [L+:W] where L is the lowest bit of the slice and W is the width of the slice. In this case, only W must be a non-negative local compile-time known numeric value. L must also be a numeric value and some architectures may require it to be compile-time known. It must also be non-negative and in-range for the type. (Section 8.7)

pacokwon:
This PR adds support for `+:` slicing.

Here's a relevant portion of the official spec as of writing:

> A slice may also be specified as [L+:W] where L is the lowest bit of the slice and W is the width of the slice. In this case, only W must be a non-negative local compile-time known numeric value. L must also be a numeric value and some architectures may require it to be compile-time known. It must also be non-negative and in-range for the type. Some architectures may allow variable indexing, in which case an out of range value will be equivalent to a shift of that amount. An index that is not compile-time known must be a bit<W> type in order to ensure that it is non-negative.

In short:
* `W` must be a non-negative local compile-time known numeric value
* If `L` is compile-time known, it must be non-negative.
* If `L` is not compile-time known, it must be a bit type.

Because `L` is not required to be compile-time known, taking `W` bits starting from bit `L` might cause out of bounds access. The following behaviors are implemented for those cases:

* Out of bound bits are zeroed on out of bounds **read operation**.
* Out of bound bits are discarded on out of bounds **write operation**.

There are not a lot of tests utilizing this syntax. Therefore, testing has also been ran on a modified version of the current test suite. The modification was a simple transformation of expressions that use the `[H:L]` slice syntax to the `[L+:W]` syntax, where `W = L + H - 1`. Positive and negative static tests, and dynamic semantics test results remain the same, aside from a few parser tests now passing.


jaehyun1ee:
Could you delete the `excludes/static/future/merged/s-1316-plus-slice.exclude` and re-run the tests?

pacokwon:
Done!

Out of the three tests in `excludes/static/future/merged/s-1316-plus-slice.exclude`, two of them are still excluded due to other reasons. The only remaining test `forloop5a.p4` passes in positive static tests.

jaehyun1ee (approved):
LGTM! Thanks

PR #140, opened 2026-03-24
Expand the IL/SL interpreter and antiunify to support struct expression to struct value assignment pattern  

sanguineman:
While writing validation rules of Wasm in P4-SpecTec, I found a pattern where struct expressions are matched against struct values, as in the following example:
```
syntax limits = {MIN i64, MAX i64?}

relation Limits_ok: context range |- limits : OK  hint(input %0 %1 %2)

rule Limits_ok/without-max:
  C_0 range |- { MIN i64, MAX eps } : OK
  -- if $le_u(i64, range)
```

In ```Limits_ok```, one of the inputs, ```limits```, is a meta-type that expects a struct value. However, the current IL/SL interpreter does not support this kind of assignment pattern.

This PR fills in that missing pattern support, and it now works correctly for such cases.

Thank you in advance for the review.


jaehyun1ee:
Closing in favor of #141

PR #145, opened 2026-03-30
feat : il semantics of eval_exp

jihongmin63:
This PR implements the semantics of `Eval_exp` and related functions in the IL interpreter using P4-SpecTec. Currently, it does not include the concrete semantics of `assign_arg, invoke_extern_func, invoke_builtin_func`, and `invoke_table_func`.

Most parts are translated straightforwardly, except for the index-related logic and invoke_defined_func.

Since P4-SpecTec requires list indices to be natural numbers, I assume that all computed indices are of type natural number. This assumption applies to `Eval_exp/idx` and `Eval_exp/slice` in 4.1-eval-exp.watsup, as well as `$access_path` and `$update_path` in 4.0-eval-exp-aux.watsup.

Additionally, the inner functions of `invoke_defined_func`—such as `do_backtrack_clause_inner, do_backtrack_clause`, and `backtrack_func`—use variables from invoke_defined_func that are not passed as arguments. To the best of my knowledge, the only way to handle this is to pass them explicitly as parameters, and the implementation in this PR follows this approach.


spec-il/4.0-eval-exp-aux.watsup
```diff
@@ -0,0 +1,259 @@
+dec $nat_sub(nat, nat) : nat
+def $nat_sub(nat_l, nat_r) = $nat_sub(nat'_l, nat'_r)
+    -- if nat'_l = $(nat_l - 1)
+    -- if nat'_r = $(nat_r - 1)
+def $nat_sub(nat_l, 0) = nat_l
+
+dec $binop_number(binop, num, num) : val
+def $binop_number(ADD, NAT nat_l, NAT nat_r) = NAT $(nat_l + nat_r)
```
jaehyun1ee: `0-stdlib.watsup` defines some useful shorthands for meta-variables, e.g., `n` for `nat` and `i` for `int`. Can we rewrite these using the shorthands, for better presentation?

spec-il/4.0-eval-exp-aux.watsup
```diff
@@ -0,0 +1,259 @@
...
+def $binop_number(ADD, INT int_l, INT int_r) = INT $(int_l + int_r)
+def $binop_number(SUB, NAT nat_l, NAT nat_r) = NAT $nat_sub(nat_l, nat_r)
+    -- if $(nat_l >= nat_r) = true
```
jaehyun1ee: Instead of defining `$nat_sub`, can we do something like

```
def $binop_number(SUB, NAT n_l, NAT n_r) = NAT n
  -- if n = $(n_l - n_r)
```

This would be elaborated to:

```
def $binop_number(SUB, NAT n_l, NAT n_r) = NAT n
  -- if i = $(n_l - n_r)
  -- if i <: nat
  -- if n = (nat) i
```

spec-il/4.0-eval-exp-aux.watsup
```diff
@@ -0,0 +1,259 @@
...
+def $upcast(C, INT, NAT nat) = NAT nat
+def $upcast(C, VAR id targ*, val) = $upcast(C, typ', val)
+    -- if (DEF tparam* `-> (ALIAS typ)) = $find_typ(C, id)
+    -- if typ' = $subst_typ($generate_map<id, typ>(tparam*, targ*), typ)
+def $upcast(C, TUP typ*, TUP val*) = TUP ($upcast(C, typ, val))*
+
+dec $downcast(ctx, typ, val) : val
+def $downcast(C, NAT, NAT nat) = NAT nat
+def $downcast(C, NAT, INT int) = NAT ($int_to_nat(int))
+    -- if $(int >= 0) = true
```
jaehyun1ee: This can be rewritten as `if $(int >= 0)`, or even shorter:

```
def $downcast(C, NAT, INT n) = NAT n
```

which implicitly means:

```
def $downcast(C, NAT, INT i) = NAT n
  -- if i <: nat
  -- if n = (nat) i
```

spec-il/4.0-eval-exp-aux.watsup
```diff
@@ -0,0 +1,259 @@
...
+
+dec $generate_map<K, V>(K*, V*) : map<K, V>
+def $generate_map<K, V>(K*, V*) = $adds_map<K, V>($empty_map<K, V>, K*, V*)
+
+dec $upcast(ctx, typ, val) : val
+def $upcast(C, INT, INT int) = INT int
+def $upcast(C, INT, NAT nat) = NAT nat
+def $upcast(C, VAR id targ*, val) = $upcast(C, typ', val)
+    -- if (DEF tparam* `-> (ALIAS typ)) = $find_typ(C, id)
+    -- if typ' = $subst_typ($generate_map<id, typ>(tparam*, targ*), typ)
```
jaehyun1ee: Is `$generate_map` necessary? How about writing:

```
-- if typ' = $subst_typ(`{ (tparam `: targ)* }, typ)
```

spec-il/4.0-eval-exp-aux.watsup
```diff
@@ -0,0 +1,259 @@
...
+def $to_int(NAT nat) = nat
+def $to_int(INT int) = int
+
+;;
+
+dec $subtyp(ctx, typ, val) : bool
+dec $subtyps(ctx, typ*, val*) : bool
+
+def $subtyps(C, typ*, val*) = $forall_($subtyp(C, typ, val)*)
+    -- if $(|typ*| = |val*|) = true
```
jaehyun1ee: Similar comment applies here. Because an if premise takes a boolean meta-expression, `-- if $(|typ*| = |val*|)` is the same.

spec-il/4.0-eval-exp-aux.watsup
```diff
@@ -0,0 +1,259 @@
...
+    -- if $find_typ(C, id) = DEF tparam* `-> deftyp
+    -- if theta = $generate_map<id, typ>(tparam*, targ*)
+    -- if deftyp = ALIAS typ
+    -- if typ_new = $subst_typ(theta, typ)
+def $subtyp(C, VAR id targ*, INJ mixop val) = $exists_(bool*)
+    -- if $find_typ(C, id) = DEF tparam* `-> deftyp
+    -- if theta = $generate_map<id, typ>(tparam*, targ*)
+    -- if deftyp = VARIANT typcase*
+    -- (if mixop_typcase typ_typcase = typcase)*
+    -- (if bool = $($(mixop = mixop_typcase) /\ $subtyps(C, ($subst_typ(theta, typ_typcase))*, val)))*
```
jaehyun1ee: `mixop = mixop_typcase` is not arithmetic, so I suppose it works without the `$` sign.

```
-- (if bool = ((mixop = mixop_typcase) /\ $subtyps(C, ($subst_typ(theta, typ_typcase))*, val)))*
```

This would work as well.

spec-il/4.1-eval-exp.watsup
```diff
@@ -6,3 +6,225 @@ rule Eval_exp/number:
...
+  
+  rule Eval_exp/binary_boolean_or:
+    C |- BIN OR (BOOL b_l) (BOOL b_r) : OK (BOOL $(b_l \/ b_r))
+  
+  rule Eval_exp/binary_boolean_impl:
+    C |- BIN IMPL (BOOL b_l) (BOOL b_r) : OK (BOOL $(b_l => b_r))
+  
+  rule Eval_exp/binary_boolean_equiv:
+    C |- BIN EQUIV (BOOL b_l) (BOOL b_r) : OK (BOOL $(b_l <=> b_r))  
+}
```
jaehyun1ee, line 33: Similar comment here. I suppose logical ops do not require the dollar sign.

spec-il/4.1-eval-exp.watsup
```diff
@@ -6,3 +6,225 @@ rule Eval_exp/number:
...
+  C |- BIN binop num_l num_r : OK $binop_number(binop, num_l, num_r)
+
+rulegroup Eval_exp/compare_boolean
+{
+  rule Eval_exp/compare_boolean_eq:
+  C |- CMP EQ (BOOL b_l) (BOOL b_r) : OK (BOOL $(b_l = b_r))
+  
+  rule Eval_exp/compare_boolean_neq:
+  C |- CMP NE (BOOL b_l) (BOOL b_r) : OK (BOOL $(b_l =/= b_r))
+}
```
jaehyun1ee: And for type-polymorphic binary ops (`=` and `=/=`) as well.

spec-il/4.1-eval-exp.watsup
```diff
@@ -6,3 +6,225 @@ rule Eval_exp/number:
...
+    -- Eval_exp: C |- exp : OK (OPT val)
+
+  rule Eval_exp/match_opt_none:
+    C |- MATCH exp NONE : OK (BOOL true)
+    -- Eval_exp: C |- exp : OK (OPT eps)
+    
+  rule Eval_exp/match_inj_else:
+    C |- MATCH exp (INJ mixop_p) : OK (BOOL false)
+    -- Eval_exp: C |- exp : OK val
+    -- if $is_inj(val) = false
```
jaehyun1ee: `-- if ~$is_inj(val)` would do.

spec-il/4.1-eval-exp.watsup
```diff
@@ -6,3 +6,225 @@ rule Eval_exp/number:
...
+  C |- LIST exp* : OK (LIST val*)
+  -- Eval_exps: C |- exp* : OK val*
+
+rule Eval_exp/cons:
+  C |- CONS exp_h exp_t : OK (LIST (val_h :: val_t*))
+  -- Eval_exp: C |- exp_h : OK val_h
+  -- Eval_exp: C |- exp_t : OK (LIST val_t*)
+
+rulegroup Eval_exp/concat 
+{
```
jaehyun1ee: Minor stylistic comment, I prefer opening curly braces at the same line as the `rulegroup`:

```
rulegroup Eval_exp/concat {
...
}
```

spec-il/4.1-eval-exp.watsup
```diff
@@ -6,3 +6,225 @@ rule Eval_exp/number:
...
+    -- Eval_exp: C |- exp_n : OK (NAT nat_n)
+}
+
+rule Eval_exp/upd:
+  C |- UPD exp_b path exp_f : OK $update_path(C, val_b, path, val_f)
+  -- Eval_exp: C |- exp_b : OK val_b
+  -- Eval_exp: C |- exp_f : OK val_f
+
+rule Eval_exp/call:
+  C |- CALL id targ* arg* : $invoke_func(C, id, targ*, arg*)
```
jaehyun1ee: Currently in P4-SpecTec, `dec` is no different from `relation`s. They are both partial functions. But considering that function call is part of the core language semantics involving several intricacies, can we instead define `$invoke_func` as a relation instead? FYI in `3-relation.watsup`, `Call_func` signature exists.

spec-il/4.1-eval-exp.watsup
```diff
@@ -6,3 +6,225 @@ rule Eval_exp/number:
...
+
+rulegroup Eval_exp/iter
+{
+  rule Eval_exp/iter_opt_some:
+    C |- ITER exp (QUEST vari) : OK (OPT val)
+    -- if $sub_opt(C, vari) = C_sub
```
jaehyun1ee: Another minor stylistic comment, I prefer the binders coming to the left of `=` as: `-- if C_sub = $sub_opt(C, vari)`.

spec-il/4.1-eval-exp.watsup
```diff
@@ -6,3 +6,225 @@ rule Eval_exp/number:
...
+
+rule Eval_exp/fail:
+  C |- exp : FAIL
+  -- otherwise
+
+;;
+
+rule Eval_exps:
+  C |- exp* : OK val*
+  -- (Eval_exp: C |- exp : OK val)*
```
jaehyun1ee: Don't we need the fail rule for `Eval_exps` as well?

spec-il/4.1-eval-exp.watsup
```diff
@@ -6,3 +6,225 @@ rule Eval_exp/number:
...
+    
+  rule Eval_exp/iter_list:
+    C |- ITER exp (STAR vari*) : OK (LIST val*)
+    -- if $sub_list(C, vari*) = C_sub*
+    -- (Eval_exp: C_sub |- exp : OK val)*
+}
+
+rule Eval_exp/fail:
+  C |- exp : FAIL
+  -- otherwise
```
jaehyun1ee, line 223: Nice use of `-- otherwise` here. It makes the specification quite concise.

spec-il/4.0-eval-exp-aux.watsup
```diff
@@ -0,0 +1,259 @@
...
+def $invoke_func_aux(`{ eps }) = $empty_map<id, typ>
+def $invoke_func_aux(`{ (id_head `: typdef_head) :: (id `: typdef)* }) = $add_map<id, typ>($invoke_func_aux(`{ (id `: typdef)* }), id_head, typ)
+    -- if typdef_head = DEF eps `-> (ALIAS typ)
+def $invoke_func_aux(`{ (id_head `: typdef_head) :: (id `: typdef)* }) = $invoke_func_aux(`{ (id `: typdef)* })
+    -- if typdef_head = PARAM \/ typdef_head = EXT
+
+dec $invoke_func(ctx, id, targ*, arg*) : res<val>
+def $invoke_func(C, id, targ*, arg*) = $invoke_func_with_values(C, id, targ_new*, val_input*)
+    -- if theta = $invoke_func_aux(C.LOCAL.TYP)
+    -- if targ_new* = $subst_typ(theta, targ)*
```
jaehyun1ee: Putting `-- if (targ_new = $subst_typ(theta, targ))*` would better show that `targ*` and `targ_new*` have the same length.

spec-il/4.0.1-eval-exp-aux.watsup
```diff
@@ -0,0 +1,259 @@
...
+def $do_backtrack_func(ctx, ctx_local, val*, eps) = FAIL
+def $do_backtrack_func(ctx, ctx_local, val*, clause_cur :: clause*) = OK val_res
+    -- if $do_backtrack_func(ctx, ctx_local, val*, clause*) = OK val_res
+def $do_backtrack_func(ctx, ctx_local, val*, clause_cur :: clause*) = $do_backtrack_clause(ctx, ctx_local, val*, clause_cur)
+    -- if $do_backtrack_func(ctx, ctx_local, val*, clause*) = FAIL
+
+dec $concat_layer(layer, layer) : layer
+def $concat_layer(layer_1, layer_2) = {TYP $extend_map<id, typdef>(layer_1.TYP, layer_2.TYP), REL $extend_map<id, reldef>(layer_1.REL, layer_2.REL), FUNC $extend_map<id, funcdef>(layer_1.FUNC, layer_2.FUNC)}
+
+dec $invoke_defined_func(ctx, id, tparam*, clause*, elsclause?, targ*, val*) : res<val>
```
jaehyun1ee: If all matches fail for `clause*` and `elsclause` exists, the `elsclause` should apply.

jaehyun1ee (commented):
Thank you for the PR. Before peeking into the mechanized semantics, I first made some minor stylistic comments. How about first working on these, to get more familiar with the SpecTec meta-language?

jihongmin63:
@jaehyun1ee The draft is finished! It currently misses the external definitions and requires refactoring; I'm planning to do it this weekend.

spec-il/4.1-eval-exp.watsup
```diff
@@ -6,3 +6,227 @@ rule Eval_exp/number:
...
+
+  rule Eval_exp/list_else:
+    C |- MATCH exp listpattern : OK (BOOL false)
+    -- Eval_exp: C |- exp : OK val
+    -- if ~$is_list(val)
+
+  rule Eval_exp/opt_else:
+    C |- MATCH exp optpattern : OK (BOOL false)
+    -- Eval_exp: C |- exp : OK val
+    -- if ~$is_opt(val)
```
jaehyun1ee: These are unnecessary (I think), because the elaboration pass guarantees that for `MATCH`, `INJ _` always operates on `INJ _` values, and similarly for `listpattern` and `optpattern`.

jaehyun1ee (changes requested):
Sorry for the late review, I made some refactoring myself, mainly inlining unnecessary meta-functions into premises and converting some meta-functions into relations. Although in the current state, relations are no different from meta-functions (they're both partial functions anyway), I believe it is a good practice to keep meta-functions *simple*, e.g., meta-functions should generally not invoke other relations from premises.

PTAL at the changes, and I left two TODOs in `4.0.3-eval-assign` and `4.4.2-eval-call-rel`. Please refactor the meta-functions with relations, likewise the changes I made.

Thanks! 

+ Let's merge this once the TODOs mentioned above are resolved :)

jihongmin63:
Dear @jaehyun1ee, the suggested changes are accepted! I am currently working on other function call such as `table_func`.

jaehyun1ee (approved):
LGTM!

PR #146, opened 2026-03-31
fix: variable shadowing

jihongmin63:
This PR modifies the part makes shadowing.

Currently, `interp.ml` uses `value_n` as the index of the slice, and the value which is inserted to the original value at the same time


jaehyun1ee:
Thanks! Nice catch. Could you make a similar patch to the SL interpreter at `p4spec/lib/interp/interp-sl/interp.ml`?

jihongmin63:
@jaehyun1ee Typos are fixed!

PR #147, opened 2026-04-02
Meta-type-checking of values given from the OCaml implementation

shyukahn:
The OCaml implementation of P4-SpecTec relies on meta-functions and relations, and also vice versa. The architecture simulators create the input values of a meta-function or a relation and provide them to the IL/SL interpreters. On a built-in or extern meta-function call, the function's OCaml implementation computes and provides the desired output value. However, these given values were no meta-type-checked, just assuming their meta-well-typedness. This PR improves this behavior, adding a guard that ensures the values are meta-well-typed. Alongside, to properly type-check function meta-values, various type definitions are extended, such as `FuncT` of `Il.typ` and functions and relations of the IL/SL interpreter context.

## Meta-type-checking

The IL/SL interpreter modules (`INTERP_IL`/`INTERP_SL`) provide functions `eval_program`, `eval_rel`, and `eval_func` that evaluate a relation or a meta-function to the architecture simulator. Within their bodies, each of them call either `check_rel_inputs` to meta-type-check the relation inputs or `check_func_inputs` to meta-type-check the function inputs.

For the output values of built-in and extern meta-functions, the bodies of the functions `invoke_extern_func` and `invoke_builtin_func` are modified to check the type of the output value.

Properly meta-type-checking the values required retaining type information in several data structures.

The function (`func.ml`) and relation (`rel.ml`) definitions in `p4spec/lib/runtime/dynamic-{il,sl}` are extended to capture the full function signature. For example, these are the IL functions and relations:

```OCaml
(* Function *)

type t =
  | Extern of tparam list * param list * typ
  | Builtin of tparam list * param list * typ
  | Table of param list * typ * tablerow list
  | Defined of tparam list * param list * typ * clause list * elseclause option

(* Relation *)

type t =
  | Extern of nottyp * Hints.Input.t
  | Defined of nottyp * Hints.Input.t * rulegroup list * elsegroup option
```

`typ` in the IL AST has a `FuncT`, but did not hold the function signature, only indicating that a meta-type is a function. It is extended to include the function signature, consisting of type parameters, the types of the parameters, and the return type.

```OCaml
(* Types *)

and typ = typ' phrase
and typ' =
  | (* omitted *)
  | FuncT of tparam list * typ list * typ (* `func` *)
```
Additionally, the `DefP` variant of `param` in the SL AST is extended to hold the function signature of the parameter.
```OCaml
(* Parameters *)

type param = param' phrase
and param' =
  | ExpP of typ * exp
  | DefP of id * tparam list * param list * typ
```

## Miscellaneous

* Meta-type related operations such as equivalence checking, subtype checking, substitution, and expansion are refactored to receive a closure (`TId.t -> Typdef.t option`) instead of a type definition environment (`TDEnv.t`), since getting the full environment from a context is expensive if it requires converting a hash table to a map.
* Type substitution of parameters is fixed to avoid variable capturing when the parameter itself is also a generic function.


jaehyun1ee (approved):
LGTM! Thanks for the PR. I made some minor refactoring to optimize the performance a bit, primarily by including input value checks into the cached code. On my machine, `make test-fast` took ~1200s with the PR before optimization, now ~1000s with optimization. (Without any meta-type-checking, it takes ~800s.)

PR #153, opened 2026-04-08
Preserve regions of P4 originated meta-values

shyukahn:
This PR extends IL/SL meta-values to include their region. The P4 lexer and parser are refactored to store the regions of meta-values that originate from a P4 program.

## Region of a meta-value

The type signature of an IL value is extended from a `note` to a `note_phrase` to store its region.

```OCaml
and value = (value', vnote) note_phrase
```

## Retaining regions from a P4 program

The P4 lexer and parsers retain regions of P4 elements when parsing them to meta-values.

The P4 lexer is refactored to store token positions in the standard lexing buffer, instead of tracking them manually. Also, it directly translates lexer positions to regions defined in `p4spec/lib/util/source.ml`. As the uses of `p4spec/lib/interface/source.ml` are all replaced, it is removed.

The P4 parser retains each symbol's position that are provided by the lexer, and includes it to the meta-value of the symbol.

## Debug premises

Given a meta-expression, debug premises evaluate it and print out the resulting value. If the region of the value is present, its location is printed together. For example, this is the result of type-checking `p4c/testdata/p4_16_samples/action_call_ebpf.p4` after adding a debug premise to print the names of control type declarations:

```
$ ./p4spectec run spec -rel Program_ok -i p4c/p4include -p p4c/testdata/p4_16_samples/action_call_ebpf.p4
spec/5-typing/5.11-typing-declaration.watsup:1088.12-1088.16: name
p4c/p4include/ebpf_model.p4:67.9-67.15: `ID filter
spec/5-typing/5.11-typing-declaration.watsup:731.12-731.16: name
p4c/testdata/p4_16_samples/action_call_ebpf.p4:28.9-28.13: `ID pipe
passed
```

## Limitations

Currently, the initial values from P4 programs retain their locations, but they lose their regions when they are operands of another expression. For example, after type-checking an `expression` value to an `expressionIR` value, the latter value will not hold the location of the former value.


jaehyun1ee (approved):
LGTM! Thanks :)

PR #155, opened 2026-04-13
Fix bug in parsing left-angle (`<`) in P4 programs

### LINKED ISSUE #43: Bug in parsing left-angle (`<`) in P4 programs
There is a bug when parsing the below program.

```p4
extern T f<T>(T x);
extern Object {
    T foo<T>();
}

struct data {
    bit<8> f;
    bit<8> foo;
}

control C(inout data d, inout bit<16> foo, Object o) {
    apply {
        if (4 + d.f < 10) { // HERE
            d.foo = (bit<8>)(o.foo<bit<16>>());
        }
    }
}
```

The `HERE` part is parsed as `4 + (d.f < 10)`, where it should actually be parsed as `(4 + d.f < 10)`. 

This correlates with the ambiguity in the P4 grammar and also the fact that `f` was declared as a generic function in the above program.

The P4 grammar is ambiguous in distinguishing identifiers from type identifiers, also mentioned in **Appendix G: P4 grammar** of the official spec as:

> The grammar is actually ambiguous, so the lexer and the parser must collaborate for parsing the language. In particular, the lexer must be able to distinguish two kinds of identifiers:
> * Type names previously introduced (TYPE_IDENTIFIER tokens)
> * Regular identifiers (IDENTIFIER token)

P4-SpecTec holds a P4 program parser implemented in OCaml. To distinguish the two tokens, the lexer is stateful, holding a set of pre-declared type identifiers and also function identifiers that *may* be generic.

A bit into the details of the P4 lexer/parser implementation,

```ocaml
let rec lexer (lexbuf:lexbuf): token = 
   match !lexer_state with
    | SIdent(id, next) ->
      begin match get_kind id with
      | TypeName true ->
        lexer_state := STemplate;
        TYPENAME
      (* the 'f' token hits here, leading the lexer state to STemplate *)
      | Ident true ->
        print_context ();
        lexer_state := STemplate;
        IDENTIFIER
      | TypeName false ->
        lexer_state := next;
        TYPENAME
      | Ident false ->
        lexer_state := next;
        IDENTIFIER
      end
```

When lexing `f` in `d.f`, the lexer is mislead into state `STemplate`. Which means, the lexer will now treat following `<` as a type specialization symbol (`L_ANGLE_ARGS`) rather than as a comparison operator (`L_ANGLE`).

I initially tried to use `%prec L_ANGLE`, as in the P4C parser, but it did not help with the situation:

```ocaml
(* we have to actually inline this into expression rule to add the %prec directive *)
| info = l_angle %prec L_ANGLE
    { Op.Lt { tags = info } }
```

Related test(s) are:

* [p4c/testdata/p4_16_samples/precedence-lt.p4](https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples/precedence-lt.p4)

shyukahn:
Fixes #43 

## Parsing left-angles (`<`) in P4 programs

Parsing left-angles in a P4 program is quite complicated. `<` can be used as a comparison operator of integers, or to specify type parameters/arguments. These usages are distinguished as different tokens, `L_ANGLE` and `L_ANGLE_ARGS`, and when the lexer reads `<`, it must determine how it is used before the parser analyzes it. This is required because the precedences of `L_ANGLE` and `L_ANGLE_ARGS` differ.

```p4
extern T f<T>(T x);
extern Object {
    T foo<T>();
}

struct data {
    bit<8> f;
    bit<8> foo;
}

control C(inout data d, inout bit<16> foo, Object o) {
    apply {
        if (4 + d.f < 10) { // HERE
            d.foo = (bit<8>)(o.foo<bit<16>>());
        }
    }
}
```
On each identifier, the parser is implemented to mark whether it requires type arguments. If so, a `<` following the identifier would be determined as `L_ANGLE_ARGS`, otherwise `L_ANGLE`. In the example above, at `HERE`, `d.f` refers to a struct field, and since its type is `bit<8>`, the following `<` is not used for type arguments.
However, due to the existence of the generic extern function `f<T>`, the lexer incorrectly determined the `f` in `d.f` as an identifier with type parameters, marking the `<` as `L_ANGLE_ARGS`. Because the operator precedence is determined as `L_ANGLE < PLUS < L_ANGLE_ARGS`, `4 + d.f < 10` was parsed as `4 + (d.f < 10)`, not `(4 + d.f) < 10`.

## Implementation

This PR fixes the issue above, mostly resembling the P4 parser implementation of p4c. The parser context in `p4spec/lib/interface/context.ml` is modified, and 2 global states are added.

```OCaml
type has_params = bool
type tid = Empty | Local of string | Global of string

type ident_kind =
  | TypeName of has_params * namespace
  | Ident of has_params * tid

and namespace = ident_kind SMap.t

type t = namespace list

(* Previously looked-up identifier *)
let previous_id : string option ref = ref None

(* Namespace of a member's parent *)
let parent_namespace : namespace option ref = ref None
```
In the context, each type name consists of its own namespace, and each identifer whose type is a type name holds a reference to it. `previous_id` is updated every time the lexer reads an identifier/type identifier, and `parent_namespace` is set to the namespace of `previous_id` when `set_parent_namespace` is called on parsing member accessing expressions/lvalues.

```OCaml
%inline memberAccessExpression:
	| e = memberAccessBase DOT set_parent_namespace m = member clear_parent_namespace %prec DOT
		{ "memberAccessBase `. member" <| [ e; m ] <<| "memberAccessExpression" <<<| (at $sloc) }
;
```
When `parent_namespace` is set, the `member` is searched only in the type's namespace. For example, `d.f` above would search `f` in `struct data`, instead of searching `f` in the current context. (This is just an intuitive explanation; the actual implementation leaves the namespace of `struct data` empty, since struct fields always do not have type parameters.)

When an extern object is declared, a namespace with its extern methods is stored within the context.

```OCaml
externObjectDeclaration:
  | al = annotationList EXTERN n = push_externName tpl = typeParameterListOpt
    L_BRACE pl = externConstructorOrMethodPrototypeList R_BRACE s = pop_scope
    { let decl = "annotationList EXTERN name typeParameterListOpt `{ externConstructorOrMethodPrototypeList }" <| [ al; n; tpl; pl ] <<| "externObjectDeclaration" <<<| (at $sloc) in
      declare_type_of_il n (has_type_params_declaration decl);
      set_type_namespace_of_il n s; (* The extern type in the context holds its namespace *)
      decl }
;
```
This is to handle cases such as `(bit<8>)(o.foo<bit<16>>())` above. `foo` is a generic method in `extern Object`, so the following `<` is determined as `L_ANGLE_ARGS`. Without this, if some parantheses are removed, e.g. `(bit<8>)o.foo<bit<16>>()`, this would fail to parse.

## Test results

`precedence-lt.p4` now passes the static type-checker test and `excludes/static/bug/left-angle-parsing.exclude` is removed.

### DIFF
--- a/excludes/static/bug/left-angle-parsing.exclude
+++ b/excludes/static/bug/left-angle-parsing.exclude
@@ -1,1 +1,0 @@
-p4c/testdata/p4_16_samples/precedence-lt.p4
--- a/p4spec/lib/interface/context.ml
+++ b/p4spec/lib/interface/context.ml
@@ -16,7 +16,7 @@
 module SMap = Map.Make (String)
 
 type has_params = bool
-type tid = string option
+type tid = Empty | Local of string | Global of string
 
 type ident_kind =
   | TypeName of has_params * namespace
@@ -55,7 +55,7 @@ let declare (id : string) (k : ident_kind) : unit =
 let declare_type id has_params = declare id (TypeName (has_params, SMap.empty))
 let declare_types types = List.iter (fun s -> declare_type s false) types
 
-let declare_var ?(tid = None) id has_params =
+let declare_var ?(tid = Empty) id has_params =
   declare id (Ident (has_params, tid))
 
 let declare_vars vars = List.iter (fun s -> declare_var s false) vars
@@ -94,7 +94,7 @@ let get_kind (id : string) : ident_kind =
     match !parent_namespace with None -> !context | Some ns -> [ ns ]
   in
   let kind =
-    match find_opt id ctx with None -> Ident (false, None) | Some k -> k
+    match find_opt id ctx with None -> Ident (false, Empty) | Some k -> k
   in
   previous_id := Some id;
   kind
@@ -129,6 +129,15 @@ let go_toplevel () =
 
 let go_local () = context := !backup
 
+let get_global_context () =
+  let rec loop c =
+    match c with
+    | [] -> failwith "ill-formed context"
+    | [ _ ] -> c
+    | _ :: l -> loop l
+  in
+  loop !context
+
 let set_type_namespace (tid : string) (ns : namespace) =
   let rec loop = function
     | [] -> []
@@ -144,9 +153,14 @@ let set_parent_namespace () =
   let ( let* ) = Option.bind in
   let namespace =
     let* parent_id = !previous_id in
-    let* _, tid_opt = find_var_opt parent_id !context in
-    let* tid = tid_opt in
-    Option.map snd (find_type_opt tid !context)
+    let* _, tid = find_var_opt parent_id !context in
+    let* tid, ctx =
+      match tid with
+      | Empty -> None
+      | Local tid -> Some (tid, !context)
+      | Global tid -> Some (tid, get_global_context ())
+    in
+    Option.map snd (find_type_opt tid ctx)
   in
   let namespace = Option.value namespace ~default:SMap.empty in
   parent_namespace := Some namespace
--- a/p4spec/lib/interface/context.ml
+++ b/p4spec/lib/interface/context.ml
@@ -16,17 +16,32 @@
 module SMap = Map.Make (String)
 
 type has_params = bool
-type ident_kind = TypeName of has_params | Ident of has_params
-type t = ident_kind SMap.t list
+type tid = string option
+
+type ident_kind =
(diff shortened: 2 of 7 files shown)


jaehyun1ee (approved):
Finally fixed! Thanks a lot 🚀

PR #157, opened 2026-04-16
Flat mixfix notation with inline arguments

KunJeong:
## Motivation
While porting #143 to downstream, I had several concerns surfaced regarding the Mixop design. I think I've come up with a rather satisfying solution , but being a substantial refactor, I wanted to try it out here as well and ask for thoughts before merging it downstream. Please read the Motivation section of the downstream PR first.

- **Original PR:** kaist-plrg/spectec-core#31

This is still a draft: only 1 of the three downstream commits are ported for now. Originally hoped to finish within the week, but there are several hurdles that we might want to discuss.

- **Runtime.Type.Typdef:** The same typedef representation is reused for elaboration, structuring and prose. However, this commit adds a `Mixop.t` to the variant case, that's only used for elaboration. The current commit does a workaround by constructing `Typdef.Defined` using `Typdef.of_deftyp` which inserts empty mixops. I think a more semantically accurate solution is having different `Typdef`s for the different passes, as even now, the `Param` and `Defining` variants are only used during elaboration.
- **Runtime.Mixop:** Due to the dependency graph, I had to place the elaboration-time mixop in `Runtime.Mixop`. Ideally it belongs in `Runtime.Static`, but `Typdef` currently depends on `Mixop`.
- **Domain.Mixop:** Similarly, `Domain.Mixop` (for IL and below) cannot be moved elsewhere because `Domain.Lib` depends on it. Again, the motivation is understood but I think it fits better in IL, as it's used in the layers after IL.

### DIFF
--- a/p4spec/lib/backend-testgen-neg/config.ml
+++ b/p4spec/lib/backend-testgen-neg/config.ml
@@ -121,7 +121,7 @@ let load_def (tdenv : TDEnv.t) (def : def) : TDEnv.t =
       let td = Type.Typdef.Extern in
       TDEnv.add id td tdenv
   | TypD (id, tparams, deftyp, _) ->
-      let td = Type.Typdef.Defined (tparams, deftyp) in
+      let td = Type.Typdef.of_deftyp tparams deftyp in
       TDEnv.add id td tdenv
   | _ -> tdenv
 
--- a/p4spec/lib/backend-testgen-neg/mutate.ml
+++ b/p4spec/lib/backend-testgen-neg/mutate.ml
@@ -64,11 +64,11 @@ and gen_from_typ' (depth : int) (tdenv : TDEnv.t) (texts : value' list)
       match td with
       | Some (Defined (tparams, td)) -> (
           let theta = List.combine tparams targs |> TDEnv.of_list in
-          match td.it with
-          | PlainT typ ->
+          match td with
+          | `Plain typ ->
               typ |> Type.Subst.subst_typ theta
               |> gen_from_typ depth tdenv texts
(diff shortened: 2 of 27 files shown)


KunJeong:
Decision is postpone for now, and land the polymorphic mixop (for the tree representation) first. Factors:
- Future backends may want to use the tree shape, e.g. for EL roundtrip. Can be circumvented if the flattening logic is proven "lossless", but this needs confirmation.
- Diverging from the Wasm codebase. Wasm-SpecTec already introduced polymorphic mixops when switching to the tree representation, so makes more sense to do that first then think about whether flattening is worth the effort.
- Carrying around the tree mixop along with Il.nottyp during elaboration is duplication of info. We could find a way to eliminate the extra mixop if flat -> tree is lossless.

KunJeong:
> * Future backends may want to use the tree shape, e.g. for EL roundtrip. Can be circumvented if the flattening logic is proven "lossless", but this needs confirmation.
- Lossy atom encoding fixed in kaist-plrg/spectec-core@3e56f2f7adb3

> * Carrying around the tree mixop along with Il.nottyp during elaboration is duplication of info. We could find a way to eliminate the extra mixop if flat -> tree is lossless.
- Tree <-> Flat mixop roundtrip implemented and elaboration simplified in kaist-plrg/spectec-core@76089f6ad4e7

> * Diverging from the Wasm codebase. Wasm-SpecTec already introduced polymorphic mixops when switching to the tree representation, so makes more sense to do that first then think about whether flattening is worth the effort.
- Divergence from Wasm is still a valid concern. We can revisit when a concrete decision is made.

PR #159, opened 2026-05-13
Replace prose pipeline with executable PL with inline annotations

KunJeong:
## Motivation

The legacy `prose` pipeline compiled SL specs to AsciiDoc through `Linearize -> Prosify -> Pl.Render`. Three constraints shaped what it could do:

- **Render-only.** The legacy `Pl` AST existed solely to feed the renderer. A PL program could not be executed against a P4 input or instrumented for coverage, so the rendered prose was untestable and semantic errors could slip through unnoticed.
- **Render hints spread across passes.** Hints harvested during elaboration (`prose`, `prose_in`, `prose_true`, `prose_false`, `prose_fields`, …) were re-derived at render time by re-walking the structuring context. The renderer carried structural decisions that belonged earlier in the pipeline.
- **Incorrect render of backtracking semantics.** `CaseI` and `TryI` rendered as two different shapes despite expressing the same backtracking semantics; branching rules came out as flat `If`/`Else` chains with no visual cue that the alternatives were fallthrough arms.

### Inline annotations
Every PL node that needs a rendering hint is wrapped in a polymorphic `Annot.t` carrying the hint record alongside the payload:
```ocaml
type hints = {
  prose         : Hints.Alter.t option;
  prose_in      : Hints.Alter.t option;
  prose_out     : Hints.Alter.t option;
  prose_true    : Hints.Alter.t option;
  prose_false   : Hints.Alter.t option;
  prose_fields  : Hints.Fields.t option;
  prose_input_exps  : Sl.exp list option;
  prose_output_exps : Sl.exp list option;
}
type 'a t = { node : 'a; hints : hints }

type exp  = ((exp', typ') note_phrase) Annot.t
type instr = ((instr', inote) note_phrase) Annot.t
type def  = (def' phrase) Annot.t
```
The SL->PL pass is renamed `annotate` to better represent its role: it walks the SL tree and attaches hints inline. The interpreter ignores the hints field; the renderer reads it. Both run on the same AST, meaning the rendered prose is always in sync with what's being executed.

### Preserved Backtracking
Faithful PL execution requires preserving the boundaries of backtracking blocks. Two refactors carry the boundary down:
- `DebugI` becomes `DebugI of exp * instr` (SL) so the following instruction nests rather than implicitly trailing. After this, single-instruction subtrees mean single execution flow and multiple arms mean backtracking — an invariant the later passes rely on.
- LL gains an explicit `TryI of block list` grouping branching arms. `GroupI` arms and branch groups that don't collapse to a `CaseI` during structuring become `TryI`s, so the renderer sees the group boundary directly.
  ```ocaml
  (* LL *)
  and instr' =
    | IfI of exp * iterexp list * block * dangle
    | CaseI of exp * case list * dangle
    | TryI of block list                  (* new *)
    | DebugI of exp
    | ...
  ```
The PL interpreter (`Interp_pl`) treats `TryI` as a backtracking block: arms run in sequence and `Backtrace (Unmatch _)` falls through to the next. Shorthand instructions are also interpreted, so shortened instructions also preserve intended semantics.

### Backtracking-block rendering
`CaseI` and `TryI` both contain backtracking semantics, with the difference that `CaseI` has explicit guards for each arm while `TryI` only has a flat instruction sequence per arm.
To render backtracking explicitly and accurately, we first render the backtracking block scopes surrounded by a `Try #N:` label. Then, each fallible instruction (`IfI`, `LetI`, `RuleI`, `CheckLetI`, `OptionGetI`) gets a "fallthrough link": either `[else #N-b]` link to its recovery target (the next arm), or `[fail #N]` linking back to the block header if on the last arm.

Now that backtracking scope is explicit, we can also flatten `IfI`s within `TryI` into `Check that`. Also drops "Else if" and renders as pure "If", because the backtracking semantics are already explicit with the labels.

Several examples.

##### Example 1 (`TryI` with `Check that`)
Before:
Screenshot 2026-05-14 at 12 01 04 AM

After:
Screenshot 2026-05-13 at 11 59 36 PM

Note the flattening into `Check that`, as well as the `a. i.` nesting because pure `TryI` don't have arms-per-guard.

##### Example 2 (`CaseI` within `CaseI`, Else rendered as If)

Before: 
Screenshot 2026-05-13 at 11 20 10 PM

After:
Screenshot 2026-05-13 at 11 20 24 PM

Note how `CaseI` resembles `TryI` but with an explicit guard. Also, the previous "Else if" is now "If", and 

The PR intentionally preserves all other rendering behavior, except for `TryI/CaseI` rendering, `Check that`, and `Else if`. The block labels are unique within relations, so that links work even across rulegroups within the same relation.

### Shorthand pass (`Pl.Shorthand`)
Recreates the shorthand pass over the new PL structure: `CheckLetI`, `OptionGetI`, `DestructI`, with a few adaptations.

- `DestructI` payload shape. The legacy `DestructI` carried bare positional targets, so the renderer had to re-derive which fields were user-visible vs. underscored at print time. The new payload threads that information through the AST itself:
  ```ocaml
  | DestructI of (string option * exp) list * exp
  ```
Each field pairs an optional source-atom name with its target binding. `Some name` is a user-visible projection the renderer prints; `None` is an underscore the renderer skips.

- `CheckLetI` carries an explicit body. `CheckLetI of exp * exp` (legacy) had no body slot, so the nested SL also introduced a bug into the shorthand pass which silently dropped the inner block it was meant to nest under the check. `CheckLetI of exp * exp * block` fixes the issue.

- Per-arm shorthand guards for `CaseI`. Two new "shorthand guards" perform the `CheckLetI` shorthand, but on `CaseI` arms:
  ```ocaml
  type guard =
    | ...
    (* Shorthands -- only emitted by the shorten pass. *)
    | CheckLetSubG of typ * exp        (* scrut <: typ, bind scrut as exp *)
    | CheckLetMatchG of pattern * exp  (* scrut matches pattern, bind scrut as exp *)
  ```
  Arms whose body opened with a sub-cast or match-rename of the scrutinee were previously shortened at render time; the pattern now lives in the AST and renders through the same logic as CheckLetI.

### SL-side expand pass (`Annotate.Expand`)
Replaces the LL-side expansion that flattened nested `CallE`s into preceding `LetI`s. The walker now operates on tree-form SL, where `LetI` nests the residual block as its body: each fresh binding is built through a deferred wrapper that closes over the residual once it is known, and extraction proceeds one call at a time so each new `LetI` correctly wraps everything after it. The pass is otherwise a direct port of the legacy logic.

## Reviewer Notes

Per-commit review is recommended, as every commit is buildable and tells a coherent story. The branch builds a parallel `pl_x` layer alongside the legacy `pl`, exercises it through its own test harness, then swaps the consumers over in the final commits.

The 12 commits map onto the sections above as follows:

| # | Commit | Section |
| - | --- | --- |
| 1 | `refactor: nest DebugI` | Backtracking preserved end-to-end |
| 2 | `refactor(prose/ll): group branching arms into TryI` | Backtracking preserved end-to-end |
| 3 | `feat(pl_x): executable prose AST, renderer, and annotation` | Inline annotations |
| 4 | `feat(interp-pl): PL interpreter and dynamic-pl runtime` | Backtracking preserved end-to-end (interpreter side) |
| 5 | `test(pl): run-pl harness and snapshots` | (test harness) |
| 6 | `feat(pl_x): shorthand pass with DestructI, CheckLetI, OptionGetI` | Adaptations & Fixes — Shorthand pass |
| 7 | `feat(pl_x): per-arm let-bind shorthand for CaseI` | Adaptations & Fixes — Per-arm shorthand guards |
| 8 | `feat(annotate): SL-side expand pass for nested calls` | Adaptations & Fixes — SL-side expand pass |
| 9 | `refactor(pl_x/render): unify CaseI and TryI as backtracking blocks` | Backtracking-block rendering |
| 10 | `refactor(annotate): replace legacy prose pipeline with pl_x` | mechanical swap — review can skip |
| 11 | `refactor(lang): rename pl_x to pl` | mechanical rename — review can skip |
| 12 | `docs: regen P4-16-spec-release.html and splice.missing` | inspect via the final rendered HTML |

- **Note:** The `CheckLetI` shorthand is enabled in this PR (the legacy version left it off) so the new code path gets exercised through the test snapshots. It can be turned off after review. 

## Future Work
- **Clean up `Annotate.Expand`.** The expansion logic is currently a direct port of the legacy LL-side logic, to keep the diff in this PR reviewable. A future dedicated PR should rewrite the logic for better readability.

## Commit Log
- refactor: nest DebugI
- refactor(prose/ll): group branching arms into TryI
- feat(pl_x): executable prose AST, renderer, and annotation
- feat(interp-pl): PL interpreter and dynamic-pl runtime
- test(pl): run-pl harness and snapshots
- feat(pl_x): shorthand pass with DestructI, CheckLetI, OptionGetI
- feat(pl_x): per-arm let-bind shorthand for CaseI
- feat(annotate): SL-side expand pass for nested calls
- refactor(pl_x/render): unify CaseI and TryI as backtracking blocks
- refactor(annotate): replace legacy prose pipeline with pl_x
- refactor(lang): rename pl_x to pl
- docs: regen P4-16-spec-release.html and splice.missing


p4spec/bin/main.ml
```diff
@@ -38,6 +35,12 @@ let runner ?(cache = true) ?(det = false) ?(arch : string option) mode
     | `SL ->
         let spec_sl = structure filenames_spec in
         (Runtime.Sim.Simulator.SL spec_sl : Runtime.Sim.Simulator.spec)
+    | `PL ->
+        let spec_pl =
+          filenames_spec |> structure |> Annotate.Expand.expand_spec
+          |> Annotate.annotate_spec |> Pl.Shorthand.shorten_defs
+        in
+        (Runtime.Sim.Simulator.PL spec_pl : Runtime.Sim.Simulator.spec)
```
jaehyun1ee, line 43: A better design is to define one API over `|> Annotate.Expand.expand_spec |> Annotate.annotate_apec |> Pl.Shorthand.shorten_defs`.

Also, I believe `Pl.Shorthand.shorten_defs` better belongs to `Annotate`.

p4spec/lib/pass/structure/struct.ml
```diff
@@ -116,9 +116,10 @@ and struct_prems' (prems_internalized : (prem * iterprem list) list)
           in
           [ instr_h ]
       | DebugPr exp ->
-          let instr_h = Ol.Ast.DebugI exp $ at in
           let instrs_t = struct_prems' prems_internalized_t instr_ret in
-          instr_h :: instrs_t
+          let instr_t = List.hd instrs_t in
```
jaehyun1ee: Ouch! `DebugI` should hold a `block` rather than an `instr`. Propagates to all def-sites of `DebugI`.

p4spec/lib/runtime/dynamic-pl/func.ml
```diff
@@ -0,0 +1,35 @@
...
+let to_string = function
+  | Extern _ -> "extern function"
+  | Builtin _ -> "builtin function"
+  | Table _ -> "table function"
+  | Defined _ -> "defined function"
+
+(* PL drops the function-type info from DefP (it carries only the id).
+   Higher-order parameters cannot be type-checked at the call boundary;
+   we substitute a placeholder type. If real specs ever use higher-order
+   params, PL's AST needs to grow the dropped fields back. *)
```
jaehyun1ee: Why bother dropping it, when it can carry it along the SL-to-PL path?

p4spec/test/test_common.ml
```diff
@@ -105,7 +106,10 @@ let frontend specdir =
 
 let elab specdir = specdir |> frontend |> Elaborate.Elab.elab_spec
 let structure specdir = specdir |> elab |> Structure.Struct.struct_spec
-let prosify specdir = specdir |> structure |> Prose.Prosify.prosify_spec
+
+let prosify specdir =
+  specdir |> structure |> Annotate.Expand.expand_spec |> Annotate.annotate_spec
+  |> Pl.Shorthand.shorten_defs
```
jaehyun1ee: Same comment as in `main.ml`.

p4spec/lib/pass/annotate/ll/free.ml
```diff
@@ -28,6 +28,8 @@ let rec free_instr (instr : instr) : t =
   | CaseI (exp, cases, _) -> free_exp exp + free_cases cases
   | OtherwiseI block -> free_block block
   | GroupI (_, _, exps, block) -> free_exps exps + free_block block
+  | TryI arms -> arms |> List.map free_block |> List.fold_left ( + ) empty
```
jaehyun1ee, line 30: I prefer `blocks` over `arms`, instead define `type arm = block` in the AST defn.

p4spec/lib/pass/annotate/annotate.ml
```diff
@@ -0,0 +1,408 @@
...
+let hints_of_func_def (ctx : Ctx.t) (id_func : id) : Annot.hints =
+  let prose_in = Ctx.find_hint_prose_in ctx (`Func id_func) in
+  let prose_true = Ctx.find_hint_prose_true ctx (`Func id_func) in
+  let prose_false = Ctx.find_hint_prose_false ctx (`Func id_func) in
+  { Annot.empty with prose_in; prose_true; prose_false }
+
+(* Hint validation *)
+
+(* `validate_hint_alter` only inspects the length of its items list, so we
+   pass a unit list of the desired arity. *)
```
jaehyun1ee: I would like these comments removed, if indeed unnecessary.

p4spec/lib/pass/annotate/linearize.ml
```diff
@@ -51,10 +57,34 @@ let rec linearize_instr (instr : instr) : Ll.Ast.block =
...
-  block |> List.concat_map linearize_instr
+  block |> List.concat_map linearize_instr |> wrap_try_arms
+
+and is_branching (instr : Ll.Ast.instr) : bool =
+  match instr.it with IfI _ | HoldI _ | CaseI _ -> true | _ -> false
+
+and wrap_try_arms (instrs : Ll.Ast.block) : Ll.Ast.block =
+  match split_leading_branches instrs with
+  | [], [] -> []
+  | [], head :: rest -> head :: wrap_try_arms rest
```
jaehyun1ee: Minor, but I prefer a better name over `head`, `rest`. Applies overall.

p4spec/lib/pass/annotate/linearize.ml
```diff
@@ -51,10 +57,34 @@ let rec linearize_instr (instr : instr) : Ll.Ast.block =
...
+  | [ single ], remainder -> single :: wrap_try_arms remainder
+  | branches, remainder ->
+      let arms = List.map (fun i -> [ i ]) branches in
+      let wrapped = Ll.Ast.TryI arms $$ (no_region, { iid = -1 }) in
+      wrapped :: wrap_try_arms remainder
+
+(* Returns (leading run of branching instructions, the rest). *)
```
jaehyun1ee: Also seems like an unnecessary comment?

p4spec/lib/pass/annotate/annotate.ml
```diff
@@ -0,0 +1,408 @@
...
+        let ctx_rel = Ctx.enter_rel ctx id in
+        let exps_pl = List.map (annotate_exp ctx_rel) exps in
+        let block_ll = Linearize.linearize_block block in
+        let block_pl = annotate_block ctx_rel block_ll in
+        let elseblock_ll_opt =
+          Option.map Linearize.linearize_block elseblock_opt
+        in
+        let elseblock_pl_opt =
+          Option.map (annotate_block ctx_rel) elseblock_ll_opt
+        in
```
jaehyun1ee: Why is the `Linearize` layer necessary in `Annot`? Seems like we can simplify the current `SL -> LL -> PL` with `SL -> PL`.

`LL` was added as a rather sloppy layer when SL was redesigned with nested `LetI` and `RuleI`, but PL was kept unchanged.

p4spec/lib/interp/interp-pl/interp.ml
```diff
@@ -0,0 +1,2035 @@
...
+                    ctx fields values
+                in
+                eval_block ctx rest
+            | _ ->
+                back_err exp_source.node.at
+                  (F.asprintf "destructure failed: %s"
+                     (Sl.Print.string_of_value ~short:true value)))
+        | _ -> (
+            let flow = eval_instr ctx instr in
+            match flow with Flow.Cont _ -> eval_block ctx rest | _ -> flow))
```
jaehyun1ee: I don't understand the distinction here. Why not define the instruction semantics entirely in `eval_instr'`, but separate them over `eval_instr'` and `eval_block`? Shouldn't it be more intuitive to call `eval_instr'` from `eval_block`, where `eval_instr'` handles all sorts of instructions?

p4spec/lib/interp/interp-pl/interp.ml
```diff
@@ -0,0 +1,2035 @@
...
+    let try_arm (arm : block) : Flow.t option =
+      try
+        match eval_block ctx arm with Flow.Cont _ -> None | flow -> Some flow
+      with Backtrace (Unmatch _) -> None
+    in
+    let results = List.filter_map try_arm arms in
+    match results with
+    | [] -> Flow.Cont []
+    | [ flow ] -> flow
+    | _ :: _ :: _ -> nondet no_region
```
jaehyun1ee: Can we be more descriptive here, as in:

```ocaml
  and eval_block_deterministic (ctx : Ctx.t) (block : block) : Flow.t =
    let eval_instr_deterministic (flow_pre : Flow.t) (instr : instr) : Flow.t =
      let open Flow in
      try
        let flow_post = eval_instr ctx instr in
        match flow_pre with
        | Cont traces_pre -> (
            match flow_post with
            | Cont traces_post -> Cont (traces_pre @ traces_post)
            | _ -> flow_post)
        | Res _ -> (
            match flow_post with
            | Cont _ -> flow_pre
            | Res _ -> nondet instr.at
            | Ret _ -> back_err instr.at "cannot have both result and return")
        | Ret _ -> (
            match flow_post with
            | Cont _ -> flow_pre
            | Res _ -> back_err instr.at "cannot have both return and result"
            | Ret _ -> nondet instr.at)
      with Backtrace (Unmatch _) -> flow_pre
    in
    try
      List.fold_left
        (fun flow_pre instr -> eval_instr_deterministic flow_pre instr)
        (Flow.Cont []) block
    with Nondet at -> back_err at "nondeterministic instruction evaluation"
```

p4spec/lib/lang/pl/ast.ml
```diff
@@ -2,210 +2,159 @@ open Util.Source
...
   | ReturnI of exp
+  | DebugI of exp
   (* Shorthands *)
-  | DestructI of (exp * string) list * exp
-  | CheckLetI of exp * exp
+  (* DestructI (fields, exp_source): at runtime, destructure [exp_source]
+     (a CaseV) positionally against [fields]. Each field is [(name_opt,
+     exp_target)]: the renderer formats visible (Some) fields and skips
+     underscore (None) ones; the interpreter assigns each [exp_target]
+     to the corresponding CaseV value regardless. *)
```
jaehyun1ee: I prefer either shortening the comment or deleting it. I suppose comments were AI-assisted, where AI tend to create verbose comment.

p4spec/lib/pass/annotate/linearize.ml
```diff
@@ -36,7 +36,13 @@ let rec linearize_instr (instr : instr) : Ll.Ast.block =
...
       [ Ll.Ast.CaseI (exp, cases_ll, dangle) $$ (at, note) ]
   | GroupI (id, rel_signature, exps_group, block) ->
-      let block_ll = linearize_block block in
+      let arms = List.map linearize_instr block in
+      let block_ll =
+        match arms with
+        | [] -> []
+        | [ single_arm ] -> single_arm
+        | _ -> [ Ll.Ast.TryI arms $$ (no_region, { iid = -1 }) ]
+      in
```
jaehyun1ee: Why doesn't this call `linearize_block`?

jaehyun1ee (changes requested):
** was edited after submitting the review

Thank you for the substantial work! And a substantial work is followed by a ... substantial review and discussion. Instead of reviewing the entire PR as a whole, I'll try reviewing this incrementally.

I had a first pass over the code, and have three big questions before having a deeper look:

1. Why should `DebugI` nest a single `instr` rather than a `block`? Dropping all tail via `List.hd` in linearize pass is dangerous, as it can alter the semantics by disregarding the trailing instructions. This can be the case when `-- debug` premise occurs in the middle of the premises:

```
rule Foo/a: ...
   -- if ...
   -- debug ...
   -- if ...
   -- if ...
```

2. Why do we need a linearization layer? LL is primarily my fault; when adding nested `LetI` and `RuleI` to SL, I did not properly propagate through PL, so added a linearization pass that simply linearizes all the nested blocks under `LetI` and `RuleI`. Relatedly, this PR is fixing the issue. Then, why should LL still be used? Can we just directly transform SL to PL?

3. Regarding the change to `CaseI`, I think it is more intuitive and precise to show this in the generated prose as "If-Else If-(Else)", rather than a "Try", since they are branching to mutually-exclusive cases. A case does not fall-through to another, thus `CaseI` does not backtrack. So I wonder what you meant by "`CaseI` and `TryI` both contain backtracking semantics."

I have a few other comments to the code, though not urgent.

And some more of thoughts in my mind just to make sure that I don't forget it completely ...

Introducing `TryI` is nice, but I also worry that it makes the spec quite difficult to read, or at least, gives impression of such. But the current backtracking nature of the IL/SL/PL semantics make `TryI` unavoidable. So seems like reducing the occurrence of `[else-#]` and `[fail-#]` is the best we can do.

This is far beyond the scope of this PR, though I see one good reason for restricting the meta-function semantics s.t. it is no longer partial. We would then validate that it is total, likewise we validate determinism in the interpreter. That way, expressions would then be total, thus we no longer need to nest `LetI` and also no longer need `[else-#] / [fail-#]` for `LetI` in PL.

KunJeong:
Thanks for the detailed response. I'll answer the high-level questions first, and address the stylistic comments one-by-one.

Overall, I realize that I should have framed the PR differently. The goal of this PR isn't to make PL executable, nor to render backtracking labels for `TryI`. Both are byproducts of the main goal: **make SL and PL semantic-preserving**. The prose diff is intentionally minimal (only changes `TryI`/backtracking rendering) to avoid mixing two separate discussions: prose readability and semantic accuracy. This PR focuses on the latter.

To put this back in context, I'll first clarify what I mean by semantic accuracy. In IL, premise failures are sensitive to the surrounding rule context: any premise failure within a rule backtracks to the *next rule*. However, the SL AST has no concept of rules. Relations hold `GroupI`s, and each `GroupI` holds multiple rules, but each rule is one or more instructions and `GroupI` holds a flattened block. We've already lost information about which instructions are rule heads (first instruction of a rule) vs which instructions are rule bodies (subsequent instructions). `GroupI`s are further jumbled up during structuring, so recovering the backtracking semantics after structuring was impossible with the original state and required a structural change.

To avoid breaking structuring altogether, this PR suggests the minimum change that can both recover backtracking semantics while staying non-intrusive to SL and structuring pass. This proceeds in three stages:
1. Make instructions fully nested, so that each top-level instruction in a `GroupI` is always a rule header.
2. Preserve backtracking semantics during linearization via wrapping sibling instructions in a `TryI`.
3. Modify the rendering of ~~`CaseI` and~~ `TryI` to honestly reflect the semantics.

- +) A more fundamental fix would be to actually preserve rule structure in the SL AST and structuring pass. This gives an additional benefit: better SL debugging, and can even enable SL -> IL lifting if done correct. But that has a much larger blast radius, well beyond our time budget. Therefore we settle for a surgical/minimal fix for now.

### 1. Why does `DebugI` nest a single instruction?

Corresponds to stage 1. The entire PR operates on top of a strong invariant, and the `DebugI` nesting both creates and uses that invariant:

> Each IL rule is represented as a **single nested instruction**, and backtracking between rules are represented as a **sequence of instructions**.

Because `LetI`, `IfI`, `RuleI` already nest their premise tail in `struct_prems`, `DebugI` was the only change we needed to ensure this property holds. Thus, the `List.hd` is actually not dangerous, as `let instrs_t = struct_prems' prems_t in` only ever returns one instruction. For instance, the example that you provided is structured as `IfI(_, DebugI(_, IfI(_, IfI(_, _))))` and nothing is dropped. However, as this was not explicit in the type level, I've updated `struct_prems'` to return a single instruction instead of a block.

Ideally, we would want to update all block-nesting instructions to nest a single instruction:
```ocaml
and instr' =
  | IfI of exp * iterexp llist * instr * dangle
  | LetI of exp * exp * iterinstr list * instr
  | RuleI of id * notexp * Hints.Input.t * iterinstr list * instr
  | DebugI of exp * instr
  | GroupI of id * rel_signature * exp list * block
```
This reveals the invariant explicitly: instruction sequences (blocks) only exist in `GroupI`. It is because of this invariant that we can wrap instruction sequences in `TryI` *after* structuring (`wrap_try_arms` in `linearize.ml`.)
I tried to scope the changes to `DebugI` because changing the type signatures of `IfI`, `LetI`, `RuleI` touches all SL call sites and adds a lot of noise to the review diff. But I'm willing to apply this too if you're on board with it.

### 2. Is `LL` necessary?
Not necessary. But `linearize.ml` and `annotate.ml` play distinct roles: linearize does structural transformation (flatten nesting, group backtracking into TryI) while annotate does pure decoration (moving hints to nodes that require them). The ~120-line cost of LL still pays for a clear separation of concerns. We can make linearization invisible to consumers by making `linearize` an internal helper of `annotate`, like you said.

### 3. Why is `CaseI` rendered like `TryI`?
**This was a logical error on my part, will roll back to the original rendering for `CaseI`.**
~~Because they are the same semantics represented as two different nodes.~~
> ~~@jaehyun1ee said:~~
> ~~I think it is more intuitive and precise to show this in the generated prose as "If-Else If-(Else)", rather than a "Try", since they are branching to mutually-exclusive cases. A case does not fall-through to another, thus CaseI does not backtrack.~~

~~One `case` of a `CaseI` has two parts: a `guard` and a `block`. Your statement applies to guards: since guards are mutually exclusive via structuring, guard matching is deterministic and we don't need backtracking to explain that. However, failure within the `block` is different: a failure within a `block` backtracks to the next `case`, which is different from a conventional If-else.~~

~~Fundamentally, both `CaseI` and `TryI` are grouped sequences of instructions. As established, all instruction sequences are initially from `GroupI`. `TryI` arms and `CaseI` cases are the post-structuring version of `GroupI` instructions.~~
~~`CaseI` is merely a slightly prettier version of `TryI`, where the first instruction is lifted to a guard, and guards are mutually exclusive with one another. This is even clearer if you look at the generated prose. Anything sequences that failed `casify` are still grouped by `TryI`, the only difference being whether the guards were lifted and grouped or not.~~

jaehyun1ee:
Short comment after reading the comment; I'll have a closer look later

https://github.com/kaist-plrg/p4-spectec/blob/7806b2fb5b009c56519db8901c52cde4b466e1e0/p4spec/lib/interp/interp-sl/interp.ml#L1353

I don't think a failure in a case block backtracks to the next case block. As can be seen in the above SL interpeter code, we *select* one case block, and evaluate that, completely disregarding the rest unmatched case blocks. So the semantics is actually "If-Else If-(Else)".

KunJeong:
You're right. I stand corrected: "failure within the block backtracks to the next case" applies for `TryI`, but only theoretically so for `CaseI`. That's because any failure within a case `block` would mean the `guard` was true, and other arms would reject the execution immediately via their `guard`s anyway. I missed that part in the reasoning. So the operational behavior of `CaseI` is indeed "If/Else If/(Else)".
That makes Try-shaped rendering more about the structural lineage (`CaseI` comes from `GroupI`), but it's not worth the extra block and labels. So I'll roll back `CaseI` rendering to the original.

jaehyun1ee:
> `GroupI`s are further jumbled up during structuring, so recovering the backtracking semantics after structuring was impossible with the original state and required a structural change.

I don't think SL loses the backtracking semantics in any way. Both IL and SL does have backtracking semantics, though their granularity are different. IL backtracking occurs on rule-granulariy. For SL, it happens at the level of `block`s. So yes, it it difficult, or impossible to recover backtracking semantics at rule-granularity in SL, but it still presents backtracking semantics as-is. We should be careful in distinguishing the two.

The presence of `GroupI` is orthogonal to the backtracking semantics represented by `block`s.

> Because `LetI`, `IfI`, `RuleI` already nest their premise tail in `struct_prems`, `DebugI` was the only change we needed to ensure this property holds. Thus, the List.hd is actually not dangerous, as `let instrs_t = struct_prems' prems_t` in only ever returns one instruction. For instance, the example that you provided is structured as `IfI(_, DebugI(_, IfI(_, IfI(_, _))))` and nothing is dropped. However, as this was not explicit in the type level, I've updated `struct_prems'` to return a single instruction instead of a block.

> Ideally, we would want to update all block-nesting instructions to nest a single instruction:

For `struct_prems` that sets-up the `Ol.instr`s corresponding to `IL.prem`s, yes. But this invariant does not hold for SL in general, and IMHO should never be. When first converting `IL.prem`s to `Ol.instr`s, the observation "Each IL rule is represented as a single nested instruction" holds, but the optimization phase in structuring can nest a sequence of instructions.

In short, I believe the invariant does not hold for SL, but it makes sense for enforcing that invariant in PL.

#### Why the invariant does not make sense in SL

For example, take two IL rules below as an example.

```
relation Foo: nat nat |- nat hint(input %0)
rule Foo/a:
   n_a n_b |- n_c
   -- if n_a = 0
   -- if n_b =/= 0
   -- let n_c = 1
rule Foo/b
  n_a n_b |- n_c
  -- if n_a = 0
  -- if $iszero(n_b)
   -- let n_c = 3
```

The `struct_prems` would indeed convert each into a single strand of instructions, and they would be concatenated after anti-unification:

```
relation Foo:
1. If (n_a = 0) then
   a. If (n_b =/= 0) then
      i. Let n_c = 1
         1. Result in n_c
2. If (n_a = 0) then
   a. If $iszero(n_b) then
      i. Let n_c = 3
         2. Result in n_c
```

Yet in structuring, we can merge two `IfI`s into one, thus merging the nested instructions. In the above example, we would merge the if condition in steps 1 and 2, thus the steps 1.a and 2.a would now form a nested sequence of instructions.

```
relation Foo:
1. If (n_a = 0) then
   a. If (n_b =/= 0) then
      i. Let n_c = 1
         1. Result in n_c
   b. If $iszero(n_b) then
      i. Let n_c = 3
         2. Result in n_c
```

This is the case where we need `IfI` nesting a `block`. You may want a closer look at:

https://github.com/kaist-plrg/p4-spectec/blob/7806b2fb5b009c56519db8901c52cde4b466e1e0/p4spec/lib/pass/structure/opt/loop/merge_if.ml#L45-L56

#### Why the invariant can make sense in PL

Overall, reading the code and comments, and moving `CaseI` out of scope, I think what we're trying to achieve with this PR is precisely: **proper representation of the backtracking semantics present in the `block` construct.** And PL makes this more explicit by introducing a `TryI of block` construct, which is intended to replace all occurrences of `block` in the AST.

With `TryI`, I think it makes more sense to refactor the PL AST as something like:

```ocaml
and instr = ((instr', inote) note_phrase) Annot.t
and instr' =
  | IfI of exp * iterexp list * instr * dangle
  | HoldI of id * notexp * iterexp list * holdcase
  | CaseI of exp * case list * dangle
  | GroupI of id * id * rel_signature * exp list * instr
  | TryI of instr list
  | LetI of exp * exp * iterinstr list
  | RuleI of id * notexp * Hints.Input.t * iterinstr list
  | ResultI of rel_signature * exp list
  | ReturnI of exp
  | DebugI of exp
  (* Shorthands *)
  | DestructI of (string option * exp) list * exp
  | CheckLetI of exp * exp * instr
  | OptionGetI of exp * exp

and holdcase =
  | BothH of instr * instr
  | HoldH of instr * dangle
  | NotHoldH of instr * dangle

and case = guard * instr
```

In this regard, relating back to the topmost comment on `GroupI`, I don't think a special treatment for `GroupI` is necessary here, just suffices to represent it as `GroupI of id * id * rel_signature * exp list * instr` instead of `GroupI of id * id * rel_signature * exp list * block`.

jaehyun1ee:
And regarding `LL`, I understand that it does have pros. I'm for preserving that layer, but making it invisible to other modules.

KunJeong:
Good points. I believe we're now on the same page conceptually. With`DebugI` nested, any `block` implies backtracking. So the original statement should be adjusted to "execution path" instead of "rule": **"Each execution path is represented as a single nested instruction, and backtracking is represented as a sequence of instructions (block)"**. The `DebugI` change was necessary for that statement to hold. Now that's clear, you're right in that `IfI` and `LetI` should nest blocks as they can gain multiple backtracking instructions during structuring, and the proposed AST design was erroneous in that regard.

My personal preference is to have backtracking explicit: something like `TryI of block` and `IfI of exp * iterexp list * instr * dangle` in the SL AST, mirroring the proposal for PL. Structuring would then explicitly construct `TryI`s whenever grouping multiple instructions. But since blocks are unambiguously backtracking, that would come down to personal preference. I prefer `TryI` because pattern-matching against `TryI` would read slightly better at call sites compared to dispatching against length of the `instr list`, and `TryI` carries semantics slightly better than `block` or `instr list`, with the cost of one more constructor depth.

I totally agree with replacing all `block` in the PL AST. And with `TryI of instr list` wrapping all block occurrences, `GroupI` also only needs `instr`. I think having SL mirror that structure would be most ideal in terms of code clarity.

jaehyun1ee:
Good! I think adding `TryI` to SL can be considered, but beyond the scope of this PR, as it would require substantial work to the optimization code in structuring.

Let's first revise the code to (i) refactor the PL AST and (ii) fix `CaseI` representation in the rendered prose. After that, I think the remaining comment would mostly be stylistic.

KunJeong:
Sounds good!

p4spec/lib/interp/interp-pl/interp.ml
KunJeong: Addressed in ab28ec1.

p4spec/lib/pass/annotate/ll/free.ml
KunJeong: Addressed in 61b05b2.

p4spec/lib/pass/annotate/annotate.ml
KunJeong: Addressed in 61b05b2.

p4spec/lib/pass/annotate/linearize.ml
KunJeong: Addressed in 61b05b2.

p4spec/lib/pass/annotate/linearize.ml
KunJeong: Addressed in 61b05b2.

p4spec/lib/lang/pl/ast.ml
KunJeong: Addressed in 61b05b2.

p4spec/bin/main.ml
KunJeong: Addressed in 3805061 and c9eab1e.

p4spec/test/test_common.ml
KunJeong: Addressed in 3805061.

p4spec/lib/runtime/dynamic-pl/func.ml
KunJeong: Good catch, fixed in 64304a9.

p4spec/lib/pass/annotate/annotate.ml
KunJeong: Linearize hidden as an internal transformation layer, 3805061.

KunJeong:
I've addressed the major decisions as well as stylistic comments and encapsulation. The remaining unresolved comments all fall under the same fundamental reason, which is somewhat unsolvable as I'll explain below.

- 000412c feat(pl/render): render CaseI as If-Else chains instead of a Try block
- 3805061 refactor(annotate): hide annotation internals
- c9eab1e refactor(annotate): move shorthand pass from lang/pl into the annotate pass
- 61b05b2 style: drop/trim unnecessary comments, rename types and variables
- ab28ec1 refactor(interp-pl): align eval_try_deterministic with SL
- 64304a9 fix(pl): re-add DefP parameters and types

---

### Why bodies stay `block` and can't all become `instr`

I tried to implement the single-`instr` body version we discussed (IfI / GroupI / CaseI arms / holdcase / CheckLetI each holding one `instr`, with `TryI` wrapping any multi-instruction body). However, it surfaced an invariant that makes the single-`instr` form unsound.

The role of `linearize` is to flatten SL's nested binders into PL's flat blocks. In SL, `LetI`/`RuleI` carry their continuation as a nested tail `block`:

```ocaml
(* sl/ast.ml *)
| LetI of exp * exp * iterinstr list * block   (* 4th arg = the tail *)
```

`linearize` prepends the head to its linearized tail, so PL's `LetI` has no continuation field:

```ocaml
(* pl/ast.ml *)
| LetI of exp * exp * iterinstr list           (* scope extends to the rest of the block *)
```

After linearization a binder and its continuation are sibling elements of one flat list; a `LetI`'s scope *is the suffix of the block it sits in*. So a body like

```
let x = f(a)
let y = g(x)
result (x, y)
```

is a three-element block that cannot either be reduced to a single `instr`, nor wrapped in `TryI` (since they are not backtrack groups).

The only way bodies genuinely become single instructions is to not flatten, which isn't ideal because we require a dedicated linearization before rendering anyways. Therefore we still need `block` for IfI/GroupI/CheckLetI, and `TryI` needs to wrap a list of blocks.

### The runtime parallel: `eval_block` vs `eval_instr'`

That's exactly what's happening with the unusual two-function split you flagged during review. `eval_instr'` handles only the **self-contained** instructions those whose semantics are complete from `ctx` alone: `IfI`, `HoldI`, `CaseI`, `GroupI`, `TryI`, `CheckLetI`, `ResultI`, `ReturnI`. The binders/sequencers `LetI`, `RuleI`, `DebugI`, `OptionGetI`, `DestructI` cannot be covered by `eval_instr'`, because their binding scope extends to the end of the block, and we can only know "the rest of the block" inside `eval_block`. So `eval_block` matches those heads and recurses into `rest`; everything else it hands to `eval_instr`, threading the `Flow`. The same thing happens in `linearize`, which you also flagged.

KunJeong:
After yesterday's discussion, I refactored both `Linearize` and `Interp-pl` further to make the two-pass role and semantic distinction clearer. The previous version of `Linearize` turned out to be a messy heuristic because it did the two passes in the wrong order (flatten-then-wrap).

### Linearization
Linearization does two things: (1) it wraps all multi-instruction sequences in `TryI`, and (2) it de-nests `LetI`, `RuleI`, and `DebugI`. (1) eliminates all blocks temporarily, and (2) converts each `instr` into a `block`. Wrap-then-flatten eliminates any ambiguity between backtracking sequences and de-nested sequences, so linearize becomes the following:

```ocaml
(* linearize.ml *)
and linearize_block (block : block) : Ll.Ast.block =
  match block with
  | [] -> []
  | [ single ] -> linearize_instr single
  | alternatives ->
      let arms = List.map linearize_instr alternatives in
      [ Ll.Ast.TryI arms $$ (no_region, { iid = -1 }) ]
```
Where `linearize_instr` is the de-nesting (2) and the TryI-wrapping (1) is now apparent. Notice that single arms are not wrapped, as they are non-backtracking and won't be rendered as a Try anyway (this way, we have better What-You-See-Is-What-You-Execute).

### Interpretation
Interpretation of de-nested instructions (`LetI`, `RuleI`, `DebugI`, and now also `OptionGetI` and `DestructI`) are different from nested instructions: they need to be supplied the rest of the block. Previously this created the split between `eval_block` and `eval_instr'`; I've merged the logic into `eval_block` which is now total. Notice this requires a thunk helper `eval_nested`, which is basically what the original `eval_instr` used to do.

```ocaml
(* interp-pl *)
and eval_block (ctx : Ctx.t) (block : block) : Flow.t =
  match block with
  | [] -> Flow.Cont []
  | instr :: block_rest -> (
      let eval_nested (eval : unit -> Flow.t) : Flow.t =
        let flow =
          try eval ()
          with Backtrace backtrace ->
            backtrace
            |> back_nest instr.node.at
                 (F.asprintf "%s failed" (Pl.Print.string_of_instr instr))
        in
        match flow with Flow.Cont _ -> eval_block ctx block_rest | _ -> flow
      in
      let iid = instr.node.note.iid in
      match instr.node.it with
      (* linearized instructions: thread block_rest for binding scope *)
      | LetI (exp_l, exp_r, iterinstrs) ->
          eval_let_instr ctx exp_l exp_r iterinstrs block_rest
      | ...

      (* nested instructions *)
      | IfI (exp_cond, iterexps, block_then, dangle) ->
          eval_nested (fun () ->
              eval_if_instr iid ctx exp_cond iterexps block_then dangle)
      | ...
```

For reference, SL's `eval_instr`:
```ocaml
(* interp-sl *)
and eval_instr (ctx : Ctx.t) (instr : instr) : Flow.t =
  Hook.on_instr instr;
  try eval_instr' ctx instr
  with Backtrace backtrace ->
    backtrace
    |> back_nest instr.at
         (F.asprintf "%s failed" (Sl.Print.string_of_instr ~short:true instr))
```

Additionally, I've examined the diff between `interp-sl` and `interp-pl` and minimized it, so you can do a cross-inspection to see what exactly changed.

p4spec/lib/pass/annotate/expand/expand.ml
jaehyun1ee, line 1: Will leave this to be reviewed in the future ...

p4spec/lib/pass/annotate/expand/transform.ml
jaehyun1ee, line 1: And this file as well ...

p4spec/lib/pass/annotate/ll/ast.ml
```diff
@@ -11,14 +11,16 @@ and instr' =
   | IfI of exp * iterexp list * block * dangle
   | HoldI of id * notexp * iterexp list * holdcase
   | CaseI of exp * case list * dangle
-  | OtherwiseI of block
```
jaehyun1ee, line 14: Deleted `OtherwiseI`, as it is no longer necessary.

p4spec/lib/pass/annotate/shorthand.ml
```diff
@@ -0,0 +1,289 @@
...
+
+   - Let exp_tmp = exp
+   - If exp_tmp matches Some:
+     - Let ?(exp_target_inner) = tmp (instr_then_h)
+     - (block_then_rest) ...
+
+   Into:
+
+   - Let exp_target_inner = ! exp
+   - block_then_rest ... *)
```
jaehyun1ee, line 154: This is *possibly* risky, because if `instr_h` (`LetI`) binds `exp_tmp`, `instr_t` (`IfI`) matches on that `exp_tmp`, and if `instrs_rest` *uses* `exp_tmp`, then after inserting `OptionGetI`, `exp_tmp` will no longer be bound in `instrs_rest`.

Leaving a safeguard for this to the (near?) future ...

p4spec/lib/lang/pl/ast.ml
```diff
@@ -35,177 +35,174 @@ type var = Sl.var
...
+  | ResultI of rel_signature * exp list
   | ReturnI of exp
+  | DebugI of exp
   (* Shorthands *)
-  | DestructI of (exp * string) list * exp
-  | CheckLetI of exp * exp
-  | OptionGetI of exp * exp
+  | DestructI of (string option * exp) list * exp
+  | CheckLetSubI of typ * exp * exp * block
+  | CheckLetMatchI of pattern * exp * exp * block
```
jaehyun1ee, line 172: Divided what was formerly `CheckLetI` into `CheckLetSubI` and `CheckLetMatchI`, mainly for performance reasons. `CheckLetSubI` is much more expensive than `CheckLetMatchI`, because it incurs an expensive runtime value-type match (via `sub`). Previously, `CheckLetI` had to perform unnecessary `sub` calls when the check was actually for `match` only, leading to significantly slow execution time.

p4spec/lib/lang/pl/ast.ml
```diff
@@ -35,177 +35,174 @@ type var = Sl.var
...
+  | OptionGetI of exp * exp * block
```
jaehyun1ee, line 173: `OptionGetI` also gains a `block`, because it is essentially a condition check plus a binding.

p4spec/lib/interp/interp-pl/interp.ml
```diff
@@ -0,0 +1,2141 @@
...
+        let value_res = Value.Make.func id tparams typs_params typ in
+        Hook.on_value value_res;
+        value_res
+
+  and eval_args (ctx : Ctx.t) (args : arg list) : value list =
+    List.map (eval_arg ctx) args
+
+  (* Instruction evaluation *)
+
+  and eval_instr (ctx : Ctx.t) (instr : instr) : Ctx.t * Flow.t =
```
jaehyun1ee, line 1044: As discussed offline, `eval_instr` now returns `Ctx.t * Flow.t` instead of `Flow.t`, so passing updated `ctx` after instructions like `LetI` / `RuleI` reads more naturally.

jaehyun1ee (approved):
Thanks for the substantial work, I added some final polishing to the code. PTAL at the comments I made, roughly summarizing what I've changed.

Will merge.

p4spec/lib/pass/annotate/shorthand.ml
KunJeong: That's a good point. Should add a check that `exp_tmp` isn't bound in `instrs_rest`.

KunJeong:
Thanks for the thorough review and comments! I'll take a closer look at the Expand pass after the smaller tests are in.

PR #163, opened 2026-06-14
Bypassing the `Value.t` abstraction in builtin/extern calls from the compiled OCaml

jaehyun1ee:
## Why

When the compiled (ML) simulator calls an extern — or an extern calls back — every argument and result used to be fully converted between the generated OCaml representation and the generic `Value.t` tree. For the Dash v1model run that meant building roughly **27.7M `Value.t` nodes**, and that conversion was the dominant cost.

This conversion is unnecessary on that path: the value leaves the compiled code and is handed straight back to compiled code. It only has to become a real `Value.t` when it is actually **stored** somewhere that gets serialized (scheduler packet context, register payloads).

This PR separates those two cases. The compiled path passes values across the boundary as-is with an O(1) cast, and pays the conversion only when a value is genuinely persisted.

**Measured:** Dash v1model STF ~12.8s → ~4.3s wall; full `@sim-ml` suite roughly halved. The IL/SL interpreters are unchanged — they still use `Value.t` end to end.

## How

A small interface, `VAL`, describes everything the extern/builtin layers need from a value (destructors, constructors, case build/match). The extern, builtin, hash, packet, and state layers are made generic over it, so the same code runs against two implementations:

- **`V_value`** — `t = Value.t`. Used by the IL/SL interpreters. Every operation is the existing `Value.t` operation; the cross-boundary operations are the identity.
- **`V_typed`** — `t = Obj.t` holding the compiled spec's native OCaml value. Used by the compiled simulator. Crossing the boundary is `Obj.repr` / `Obj.obj`.

The interface has two distinct cross-boundary operations, which is the heart of the change:

```ocaml
(* Pass-through: value goes straight back to compiled code, never read here. *)
val to_value : t -> Value.t
val of_value : Value.t -> t

(* Convert: value is about to be stored in a serialized field, so it must
   become a real Value.t of its (statically known) spec type. *)
```

Under `V_value` both pairs are the identity. Under `V_typed`, `to_value`/`of_value` are casts (the fast path), and `marshal`/`unmarshal` do a real per-type conversion (the only place the old cost is still paid, and only when something is being stored).

## Concrete code changes

**1. Generated extern dispatch — convert → cast.** Inputs and outputs of compiled extern calls no longer go through `unmarshal_*` / `marshal_*`:

```ocaml
(* before: deep-convert each argument *)
let a_i = unmarshal_<typ> (List.nth args__ i)
...

(* after: cast in, cast out *)
let a_i = (Obj.magic (List.nth args__ i) : <ocaml typ>)
...
```

The extern wrappers in `rel.ml` / `func.ml` change the same way, which is what drops the two costly conversions (`marshal_eval_context`, `marshal_arch`) from the hot path.

**2. The program is converted exactly once.** `eval_program` receives a real parsed `Value.t`, so it `unmarshal`s it to a `p4program` a single time, then runs the typed path.

**3. Two parallel stacks, one chosen per run** (`make.ml`). A `V_value` stack for the IL/SL interpreters and a `V_typed` stack for the compiled simulator; `init_mode` wires up only the active one.

**4. Case build/match now take the value's spec type.** `Value.t` carries its own constructor tag at runtime, but a bare `Obj.t` is type-erased, so `V_typed` needs the spec type to pick the right OCaml projection. This also removed some hardcoded type-name guesses that previously mis-cast values.


p4spec/lib/backend-ocaml/compiled_stub/ctx.ml
```diff
@@ -12,13 +12,13 @@ module Typ = Runtime.Type.Typ
...
 
 type iface__ = {
-  call_builtin  : (Value.t -> unit) -> Domain.Lib.Id.t -> Typ.t list -> Value.t list -> Value.t;
   parse_program : string list -> string list -> Run.parse_result;
 }
 
 type extern__ = {
   eval_extern_rel  : string -> Value.t list -> Run.rel_result;
   eval_extern_func : string -> Typ.t list -> Value.t list -> Run.func_result;
+  call_builtin     : (Value.t -> unit) -> Domain.Lib.Id.t -> Typ.t list -> Value.t list -> Value.t;
```
jaehyun1ee: This should not be in `extern__`, instead should belong in `iface__`.

p4spec/lib/backend-sim/spec_impl/unpack.ml
```diff
@@ -1,98 +1,111 @@
...
+
+  let unpack_p4_precision_numberValue (value : vt) : Bigint.t * Bigint.t =
+    (* Dispatch on the actual [value] constructor. [V_native]'s [( |>> )] projects
+       by arity WITHOUT validating the constructor, so the previous
+       [try unpack_p4_fixedBit with _ -> ...] fallback is dead under [V_native] and
+       silently mis-reads e.g. a variableBit ("nat `. nat V int", 3 args) as a
+       fixedBit ("nat W int", 2 args), yielding the wrong (width, value). *)
+    let mixop, _ = Mixfix.split (V.Get.case value Typs.value) in
+    let canon s = Mixop.string_of_mixop (Value.Mixops.of_string s) in
+    let m = Mixop.string_of_mixop mixop in
```
jaehyun1ee, line 59: Not a good idea to stringify a mixop. Try structural eq or something else.

p4spec/lib/backend-sim/typs.ml
```diff
@@ -0,0 +1,31 @@
...
+let value = var "value"
+let transition_result = var "transitionResult"
+
+(* State-persist types, passed to [V.marshal]/[V.unmarshal] so the typed
+   [Obj.t] becomes a real [Value.t] before it is yojson-serialized. *)
+let eval_context = var "evalContext"
+let type_ir = var "typeIR"
+
+(* The lone mixop threaded structurally (the [( |>>? )] REJECT test). *)
+let mo_reject_error_value = Value.Mixops.of_string "REJECT errorValue"
```
jaehyun1ee, line 19: Placement issue

p4spec/lib/backend-sim/v1model/object.ml
```diff
@@ -19,25 +23,34 @@ module Make (Spec : Spec.S) = struct
...
+
+    let bytes_pair_of_yojson :
+        Yojson.Safe.t -> (Bigint.t * Bigint.t, string) result = function
+      | `List [ jp; jb ] -> (
+          match
+            (Util.Json.bigint_of_yojson jp, Util.Json.bigint_of_yojson jb)
+          with
+          | Ok packets, Ok bytes -> Ok (packets, bytes)
+          | Error e, _ | _, Error e -> Error e)
+      | _ -> Error "expected a counter (packets, bytes) pair"
```
jaehyun1ee: Can we do something about it? Maybe a polymorphic helper in json utils?

p4spec/lib/runtime/valrep/valrep.ml
```diff
@@ -0,0 +1,11 @@
+(* Re-export of the boundary contracts and the interpreter instance so consumers
+   keep using [Valrep.SAFE] / [Valrep.RAW] / [Valrep.VAL] and [Valrep.V_value].
+   The contracts live in [sig.ml] and the [V_value] instance in [val_value.ml];
+   [V_native] (the native-OCaml instance) lives in [backend-ocaml], where the
+   generated symbols are in scope. *)
+
+module type SAFE = Sig.SAFE
+module type RAW = Sig.RAW
```
jaehyun1ee: How about we rename this to `UNSAFE`?

p4spec/lib/runtime/prof.ml
jaehyun1ee, line 1: Get rid of this file!

README.md
```diff
@@ -118,11 +118,9 @@ $ docker build -f p4spectec_p4c.dockerfile -t p4spectec_with_p4c:latest .
...
 ## P4-SpecTec: A language specification framework for P4
 
-The spec source files are located in the `spec/p4` directory. A compile-optimized
-variant lives in `spec/p4-comp`, which shares `spec/p4` via symlinks and overrides
-only the stdlib map functions (defined-function versions for the ML compiler; see
-`docs/superpowers/specs/2026-06-12-dual-spec-symlink-design.md`). Use `spec/p4` for
-the interpreter and `spec/p4-comp` when compiling.
+The spec source files are located in the `spec` directory. The same tree is used
+by both the interpreter and the ML compiler; stdlib map functions are native
+builtins under the VAL abstraction, so no compile-optimized variant is needed.
```
jaehyun1ee, line 123: Don't mention the ML backend at all.

p4spec/lib/backend-sim/psa/object.ml
```diff
@@ -1,37 +1,52 @@
...
+
+    let bytes_pair_of_yojson :
+        Yojson.Safe.t -> (Bigint.t * Bigint.t, string) result = function
+      | `List [ jp; jb ] -> (
+          match
+            (Util.Json.bigint_of_yojson jp, Util.Json.bigint_of_yojson jb)
+          with
+          | Ok packets, Ok bytes -> Ok (packets, bytes)
+          | Error e, _ | _, Error e -> Error e)
+      | _ -> Error "expected a counter (packets, bytes) pair"
```
jaehyun1ee: Also for this as well.

p4spec/lib/backend-sim/psa/object.ml
```diff
@@ -107,7 +116,9 @@ module Make (Spec_Func : Spec.Func.S) = struct
     (* Type *)
 
     type color = RED | GREEN | YELLOW [@@deriving yojson]
-    type t = Packets of color list | Bytes of color list [@@deriving yojson]
+
+    (* Flat [array] indexed by meter index. *)
```
jaehyun1ee, line 111: Delete this comment.

p4spec/lib/interface/builtin/call.ml
```diff
@@ -2,109 +2,118 @@ module Fresh_ = Fresh
...
-module Value = Runtime.Value
-module Run = Runtime.Dynamic_Runner.Signature
 open Error
 open Util.Source
 
-(* Extensibility point: extra or override builtins per interface *)
+(* Shared builtin state counter *)
 
-type impl = (Value.t -> unit) -> region -> Typ.t list -> Value.t list -> Value.t
+let ctr : int ref = ref 0
```
jaehyun1ee, line 11: Don't share this, push it into `Make_funcs`.

p4spec/lib/interface/builtin/sets.ml
```diff
@@ -3,101 +3,113 @@ open Lang
...
+
+  let compare_v (typ : typ) (v_a : V.t) (v_b : V.t) : int =
+    Value.compare (V.marshal typ v_a) (V.marshal typ v_b)
+
+  let norm cmp xs = List.sort_uniq cmp xs
+  let mem cmp x ys = List.exists (fun y -> cmp x y = 0) ys
+  let inter cmp a b = List.filter (fun x -> mem cmp x b) a
+  let union cmp a b = norm cmp (a @ b)
+  let diff cmp a b = List.filter (fun x -> not (mem cmp x b)) a
+  let subset cmp a b = List.for_all (fun x -> mem cmp x b) a
```
jaehyun1ee, line 19: Find deadcode and remove it.

p4spec/lib/backend-sim/make.ml
jaehyun1ee, line 1: Do something with the comments.

p4spec/lib/backend-sim/typs.ml
jaehyun1ee, line 1: Is this *really* necessary? If it isn't performance-critical, I prefer not having it?

p4spec/lib/pass/compile/gen/dispatch.ml
jaehyun1ee, line 1: Do something about the comments. But maybe not in this PR, perhaps?

p4spec/lib/pass/compile/gen/common.ml
jaehyun1ee, line 1: Also do something about the comments, but in another PR?

p4spec/lib/pass/compile/gen/func.ml
```diff
@@ -62,10 +62,22 @@ let compile_params (ctx : Ctx.t) (params : param list) :
...
-(* Extern functions *)
+(* Extern + builtin functions.
 
-let compile_extern_func (ctx : Ctx.t) (reverse_dispatch : reverse_dispatch)
-    (id : id) (params : param list) (typ_ret : typ) : Ctx.t * Ml.funcdef list =
+   Both cross a [Value.t]-typed runtime boundary that, under the compiled (ML)
+   path, actually carries typed [Obj.t] values. So both [Obj.magic] their args
+   in and their result out instead of converting. The only per-kind differences
+   are the call field, its extra leading args, and how the result/error is
+   shaped — captured by [make_result]. *)
```
jaehyun1ee: Think of a less verbose comment.

p4spec/lib/pass/compile/template/functor.ml
```diff
@@ -13,6 +13,12 @@ let eval_program (relname__ : string) (includes__ : string list)
     (path__ : string) : Run.program_result =
   match (!cur__).iface.parse_program includes__ [path__] with
   | Run.Pass value_program -> (
+      (* The parsed program is a real [Value.t] from the parser, but [eval_rel]
+         [Obj.magic]-casts its inputs. So unmarshal it once here into the typed
+         [p4program], then [Obj.repr] it. *)
```
jaehyun1ee: Comment issue.

p4spec/lib/pass/compile/gen/rel.ml
```diff
@@ -89,33 +89,36 @@ let compile_extern_rel (ctx : Ctx.t) (id : id)
         ("param__" ^ string_of_int i, Some typ_ml))
       typs_input
   in
-  (* Marshal inputs *)
+  (* Pass inputs to the extern as typed [Obj.t] through the [Value.t list]
+     boundary ([Obj.magic]) instead of converting — this drops the
+     [marshal_eval_context]/[marshal_arch] that dominated. *)
```
jaehyun1ee: Comment issue.

p4spec/lib/pass/compile/codegen.ml
jaehyun1ee, line 1: Also comment issue.

p4spec/lib/runtime/dynamic-runner/signature.ml
```diff
@@ -29,9 +29,28 @@ module type CACHE = sig
...
+  val init : unit -> unit
+  val checkpoint : unit -> int
+  val seff : int -> int -> bool
+end
+
 (* Interface for the interaction between SpecTec and the defined language *)
 
 module type INTERFACE = sig
+  (* The interface's builtins at any value rep; [call_builtin] is
+     [Builtins(Valrep.V_value).invoke]. *)
```
jaehyun1ee, line 47: Comment issue

p4spec/lib/runtime/dynamic-runner/signature.ml
```diff
@@ -29,9 +29,28 @@ module type CACHE = sig
   val cache_off : unit -> unit
 end
 
+(* Builtins, at an arbitrary value representation [V]. An interface owns its
+   builtin set + extensions; exposing them as a functor over [V] lets a consumer
+   (e.g. the simulator's native ML path) obtain the same builtins at a second
+   rep without reconstructing them from the builtin library internals. *)
```
jaehyun1ee: Comment issue

p4spec/lib/runtime/valrep/sig.ml
jaehyun1ee, line 1: Comment issue

jaehyun1ee (commented):
Comments for myself

jaehyun1ee:
Superseded by #189 (recreated with renamed branch after this PR's head branch was renamed). Continuing there.

PR #165, opened 2026-06-26
Track test inputs in dune so snapshots don't go stale

KunJeong:
### Motivation
Snapshot test rules read {spec,p4c,excludes,patches,testdata} from the repo root outside the dune project (at p4spec/), so dune tracked none of them as dependencies. Editing the spec, excludes, or a P4 sample left the test stale until a manual `make clean`.

### Core Concept
This PR moves the dune workspace root to the repo root. This is done by a new `dune-project` and dune ((dirs)/(data_only_dirs)) at the root, bringing spec/ and P4 testdata into the workspace without relocating them. Each snapshot rule now declares its inputs via (deps (source_tree ...)), so dune invalidate test cache if and only if any inputs change (i.e. no manual `make clean` required).

### Scope
- New root `dune-project` and `dune`.
- `(deps (source_tree ...))` on every snapshot rule in `test/{lang,run,sim,parse}/dune`.
- `_build` now lives at the repo root → `Makefile` exe paths and `ci-main.yml` updated to match.
- Exclude-path prefix in `test/{run,sim,parse}/test.ml` corrected `../../../../../ → ../../../` to match the shallower discovered paths after the move.
- Path strings in all `.expected` files are moved from `../../../../../ → ../../../`.


KunJeong:
I've rebased onto `concrete-refactor-prose` and changed the base branch to `concrete-refactor-prose`, so that it can be merged immediately once #159 is in.

jaehyun1ee (approved):
Thanks!

PR #166, opened 2026-06-27
Add a fast @micro test over a small p4 corpus

KunJeong:
## Motivation

The current test suite is thorough but slow: `@run` and `@sim` iterate over the full p4c and p4testgen suite (hundreds of programs each), so there is no quick way to check that a change hasn't broken something across the pipeline. This PR adds a fast tier intended as a development sanity check, not as a replacement for the full suite.

## Core Concept

A small hand-written P4 corpus (16 programs) exercised through every test category the suite has:

- **parse** — P4 → EL, including one syntax-error input
- **run** — typing (`Program_inst`, `Program_ok`)
- **sim** — `v1model`, `ebpf`, `psa` execution

`@micro` runs the corpus under all three interpreters (`il`, `sl`, `pl`) with non-deterministic reduction (`choose_sequential`); parse, being reduction-independent, lives here too. `@micro-det` re-runs the run/sim programs under deterministic reduction (`-det`, `choose_deterministic`), diffed against the same expectations. `make test-micro` builds both, plus `@speclang` for the corpus-free specification pipeline (`elab`, `struct`, `annotate`; EL → IL → SL → PL). A full run finishes in a few seconds.

## Scope

- `p4spec/test/micro/dune`: parse/run/sim rules and the `@micro` (seq) and `@micro-det` (det) aliases.
- Curated corpus: `programs/` (11, positive typing), `programs-neg/` (1, negative typing), `programs-parse-neg/` (1, syntax error), `sim-{v1model,ebpf,psa}/` (`.p4` + `.stf`).
- `Makefile`: `test-micro` (`@speclang @micro @micro-det`), `test-micro-det`.
- `.expected` snapshots for all tests.

## Corpus design

Each program is small and targets a distinct cluster of constructs:

- `programs/` (typing): casts, control/action/`for`, declarations with structs and tuples, expressions, functions and generics, header stacks, overloading, parsers with `select` and stack `.next`, tables with `switch`, and enums/typedefs/header-unions.
- `programs-neg/`, `programs-parse-neg/`: one ill-typed and one ill-formed program.
- `sim-{v1model,ebpf,psa}/`: each self-contained with a non-empty `.stf` packet trace so the architecture actually runs; the v1model program is the richest, exercising `verify/update_checksum`, `hash`, `mark_to_drop`, a table, and a `for` loop.

(Test corpus was authored with the assistance of Claude Opus 4.8)

Two properties keep it trustworthy despite its size: it runs the same test executables and the same specification as the full suite (no new driver), and each run/sim program is checked under both `seq` and `-det` mode, so agreement confirms the spec is unambiguous.

## Future work

- Extend the corpus reactively: when a regression escapes the tier, add the minimal program that would have caught it.


KunJeong:
Note: this PR is stacked onto #165.

jaehyun1ee:
Thanks for the addition, a fast test suite was necessary for faster turn-around time. I expect that the corpus will grow over time, so given this expectation, seems like reusing parts of the official p4c test suite might be a better direction. I made a similar PR at #173, PTAL.

jaehyun1ee:
Merging in favor of a less-intrusive approach to micro-tests.

PR #167, opened 2026-06-30
Split elaboration into two separate passes: EL → IL and IL → AL

shyukahn:
## Motivation

P4-SpecTec is a framework that originated from Wasm-SpecTec. They both provide a specification language that is parsed into EL and perform an EL to IL elaboration pass. Although they share the terms 'IL' (Internal Language) and 'elaboration', their implementations differ significantly.

[Wasm-SpecTec's IL](https://github.com/WebAssembly/spec/blob/main/spectec/doc/IL.md) is a more explicit, declarative representation of a type-checked specification. Its elaboration pass type-checks EL and translates it into IL.

P4-SpecTec's IL, on the other hand, is _algorithmic_. To validate that a specification is algorithmic, P4-SpecTec's elaboration additionally performs binding analysis during the translation from EL to IL.

This PR aims to narrow this gap between P4-SpecTec and Wasm-SpecTec by refactoring the elaboration pass into two separate passes.

## IL and AL

To split elaboration into two passes, IL is renamed to **AL** (Algorithmic Language), and a new declarative IL is defined. A new elaboration pass translates EL to the new IL, and another pass, named (temporarily) algo, translates the new IL to AL. From here on, the terms 'IL' and 'elaboration' refer to the new IL and new elaboration, respectively.

The IL AST and AL AST differ in the type definitions of rulegroups and table rows.

```OCaml
(* IL *)

type rule = rule' phrase
and rule' = id * notexp * prem list

and rulegroup = rulegroup' phrase
and rulegroup' = id * rule list

and elsegroup = elsegroup' phrase
and elsegroup' = id * rule

and tablerow = tablerow' phrase
and tablerow' = arg list * exp

(* AL *)

type rulematch = exp list * exp list * prem list
and rulepath = id * prem list * exp list

and rulegroup = rulegroup' phrase
and rulegroup' = id * rulematch * rulepath list

and elsegroup = elsegroup' phrase
and elsegroup' = id * rulematch * rulepath

and tablerow = tablerow' phrase
and tablerow' = exp list * arg list * exp * prem list
```

## Passes

The elaboration pass converts EL to IL and performs type-checking, including dimension analysis. Most of its implementation is reused, except for dimension analysis. The previous implementation relied on the result of binding analysis, which is now performed after elaboration. Instead, Wasm-SpecTec's approach is adopted. A detailed explanation is given in #161.

The algo pass converts IL to AL and performs binding analysis and sidecondition insertion. Binding analysis is implemented to handle iterators in a more sophisticated manner, since dimension analysis is now done beforehand. Sidecondition insertion operates on AL and reuses the existing implementation.

### Binding analysis

During binding analysis, an `Iterctx` is threaded through the code to keep track of the iterators and the iterating variables.

```OCaml
(* Iterctx *)
type t = (iter * var list * var list) list
```

Each entry of an `Iterctx` records an iterator together with two lists of variables: the bound variables and the binding variables. `Iterctx` is also used to construct iterating premises and expressions. For example, consider the following code that analyzes let premises and iterating premises:

```OCaml
and analyze_let_prem (ctx : Ctx.t) (at : region) (iterctx : Iterctx.t)
    (exp_l : exp) (binds_l : BEnv.t) (exp_r : exp) :
    Ctx.t * VEnv.t * prem * prem list =
  (* Omitted... *)
  let iterctx =
    iterctx
    |> Iterctx.filter_bound (fun id typ iters ->
           VEnv.find_opt id venv_r
           |> Option.map (fun (typ_r, iters_r) ->
                  Typdim.sub (typ_r, iters_r) (typ, iters))
           |> Option.value ~default:false)
    |> Iterctx.add_vars_bind venv_l
  in
  Iterctx.validate at iterctx;
  let prem = Iterctx.iterate_prem iterctx prem in
  (ctx, venv, prem, prems)

and analyze_iter_prem (ctx : Ctx.t) (iterctx : Iterctx.t) (prem : prem)
    (iter : iter) (vars : var list) : Ctx.t * VEnv.t * prem * prem list =
  let iterctx = (iter, vars, []) :: iterctx in
  analyze_prem ctx iterctx prem
```

`analyze_iter_prem` updates the `iterctx` and passes it to the inner premise. `analyze_let_prem` makes a new `iterctx` by adjusting the bound and binding variables according to the analysis result, and then builds an iterating premise from `iterctx` if it is nonempty.

Sideconditions generated from renaming multi-binding occurrences and desugaring partial bindings also exploit `iterctx`, so that the inserted premises are iterated to the correct dimension.

## Miscellaneous

The following command prints the specification in AL after the algo pass:
```bash
$ ./p4spectec algo spec
```

As IL is renamed to AL, many files and tests referring to the IL interpreter were renamed to include AL instead of IL.


p4spec/lib/backend-sim/ebpf/pipe.ml
```diff
@@ -7,7 +7,7 @@ module Sim = Runtime.Sim.Simulator
 open Util.Source
 open Error
 
-module Make (Interp_IL : Sim.INTERP_IL) (Interp_SL : Sim.INTERP_SL) : Sim.ARCH =
+module Make (Interp_AL : Sim.INTERP_AL) (Interp_SL : Sim.INTERP_SL) : Sim.ARCH =
```
jaehyun1ee: [Note] A separate issue, which I think relates to the recent PR #159, commenting as a note for myself.

p4spec/lib/backend-sim/psa/pipe.ml
```diff
@@ -9,7 +9,7 @@ open State
 open Error
 open Util.Source
 
-module Make (Interp_IL : Sim.INTERP_IL) (Interp_SL : Sim.INTERP_SL) : Sim.ARCH =
+module Make (Interp_AL : Sim.INTERP_AL) (Interp_SL : Sim.INTERP_SL) : Sim.ARCH =
```
jaehyun1ee: [Note] This as well.

p4spec/lib/backend-sim/v1model/pipe.ml
```diff
@@ -9,7 +9,7 @@ open State
 open Error
 open Util.Source
 
-module Make (Interp_IL : Sim.INTERP_IL) (Interp_SL : Sim.INTERP_SL) : Sim.ARCH =
+module Make (Interp_AL : Sim.INTERP_AL) (Interp_SL : Sim.INTERP_SL) : Sim.ARCH =
```
jaehyun1ee: [Note] And this.

p4spec/lib/backend-sim/driver.ml
```diff
@@ -9,12 +9,12 @@ open Util.Source
...
 module Make
-    (MakeArch : functor (Interp_IL : INTERP_IL) (Interp_SL : INTERP_SL) -> ARCH)
-    (MakeInterp_IL : functor (Arch : ARCH) -> INTERP_IL)
+    (MakeArch : functor (Interp_AL : INTERP_AL) (Interp_SL : INTERP_SL) -> ARCH)
+    (MakeInterp_AL : functor (Arch : ARCH) -> INTERP_AL)
     (MakeInterp_SL : functor (Arch : ARCH) -> INTERP_SL)
     (MakeInterp_PL : functor (Arch : ARCH) -> INTERP_PL) : DRIVER = struct
-  module rec Arch : ARCH = MakeArch (Interp_IL) (Interp_SL)
-  and Interp_IL : INTERP_IL = MakeInterp_IL (Arch)
+  module rec Arch : ARCH = MakeArch (Interp_AL) (Interp_SL)
```
jaehyun1ee: [Note] And this as well, add `(Interp_PL)` to it?

p4spec/lib/backend-sim/placeholder.ml
```diff
@@ -5,7 +5,7 @@ module Sim = Runtime.Sim.Simulator
 open Error
 open Util.Source
 
-module Make (Interp_IL : Sim.INTERP_IL) (Interp_SL : Sim.INTERP_SL) : Sim.ARCH =
+module Make (Interp_AL : Sim.INTERP_AL) (Interp_SL : Sim.INTERP_SL) : Sim.ARCH =
```
jaehyun1ee: [Note] And this.

p4spec/lib/pass/algo/binding/bind.ml
```diff
@@ -6,8 +6,6 @@ open Error
 open Envs
 open Util.Source
 
-(* Binding occurrences of identifiers, singular or multiple (parallel) *)
```
jaehyun1ee, line 9: Can we resurrect this comment, if it is not stale?

p4spec/lib/pass/algo/binding/bind.ml
```diff
@@ -19,8 +17,6 @@ module Occ = struct
     | Multi typ -> Multi (Typdim.add_iter iter typ)
 end
 
-(* Environment for identifier bindings *)
```
jaehyun1ee, line 22: Can we resurrect this comment, if it is not stale?

p4spec/lib/pass/elaborate/dataflow/collectbind.ml
```diff
@@ -5,159 +5,143 @@ open Error
 open Runtime.Static.Envs
 open Util.Source
 
-(* Collect binding identifiers,
-   while enforcing the invariant that binding identifiers
-   can only occur in invertible constructs *)
```
jaehyun1ee, line 10: Can we resurrect this comment, if it is not stale?

p4spec/lib/pass/algo/binding/ctx.ml
```diff
@@ -0,0 +1,75 @@
...
+
+type t = {
+  (* Free identifiers over the entire definition *)
+  frees : IdSet.t;
+  (* Bound variables so far *)
+  bounds : VEnv.t;
+  (* Metavariables so far *)
+  menv : MEnv.t;
+  (* Typedefs so far *)
+  tdenv : TDEnv.t;
```
jaehyun1ee, line 17: (Note) I'll do the refactoring later, again as a note for myself. Reorder and rename:

```ocaml
frees : IdSet.t;
venv : VEnv.t;
tdenv : TDEnv.t;
menv : MEnv.t;
```

p4spec/lib/pass/algo/binding/dimension.ml
```diff
@@ -0,0 +1,77 @@
...
+  let typ', at = (exp.note, exp.at) in
+  match exp.it with
+  | BoolE _ | NumE _ | TextE _ -> venv
+  | VarE id -> (
+      let typ = typ' $ at in
+      match VEnv.find_opt id venv with
+      | None -> VEnv.add id (typ, iters) venv
+      | Some (typ_prev, iters_prev) ->
+          if Typdim.sub (typ, iters) (typ_prev, iters_prev) then
+            VEnv.add id (typ, iters) venv
```
jaehyun1ee, line 23: What is happening here? Can you explain at a high-level?

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1663,18 +1493,13 @@ let elab_clause (ctx : Ctx.t) (at : region) (id : id) (tparams : tparam list)
...
   let ctx_local = Ctx.add_tparams ctx_local tparams in
-  let ctx_local, args_il, sideconditions_il =
-    elab_clause_input_with_bind ctx_local at params_il args
-  in
-  let ctx_local, prems_internal = elab_prems_with_bind ctx_local prems in
-  let sideconditions_internal = List.map internalize_prem sideconditions_il in
-  let prems_internal = sideconditions_internal @ prems_internal in
+  let ctx_local, args_il = elab_args ~as_def:true at ctx_local params_il args in
+  let ctx_local, prems_internal = elab_prems ctx_local prems in
   check_prems_internal at prems_internal;
```
jaehyun1ee, line 1437: This checks

```ocaml
  if List.length prems_internal = List.length prems_non_else_internal then ()
  else if List.length prems_internal = List.length prems_non_else_internal + 1
  then
    check
      (List.for_all is_pure_prem_internal prems_non_else_internal)
      at "cannot have non-pure premises alongside an otherwise premise"
  else error at "cannot use multiple otherwise premises"
```

But given that `elab_prems` do *not* emit side-conditions anymore in elaboration, I believe that `check_prems_internal` should somehow be patched as well. Maybe move the check over to the `algo` pass?

e.g.,

```
relation Foo: int : int (hint input %0)
rule Foo/a: 1 : 43
rule Foo/b: 0 : 42
  -- otherwise   
```

Here, if we were to use the old elab pass, then the input `0` in rule `b` would be converted to a `-- if x = 0`, then the `-- otherwise` becomes invalid because we use it together as `-- if x = 0 -- otherwise`.

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -2158,3 +1963,4 @@ let elab_spec (spec : spec) : Il.spec =
   let ctx, spec_il = elab_defs ctx spec in
   populate_typs ctx;
   spec_il |> populate_rules ctx |> populate_clauses ctx
+  |> Dimension.analyze_spec
```
jaehyun1ee, line 1792: Nice!

p4spec/lib/pass/algo/binding/iterctx.ml
```diff
@@ -0,0 +1,90 @@
...
+  in
+  iterctx
+
+let add_var_bind (id : Id.t) (typ : typ) (iters : iter list) (iterctx : t) : t =
+  let venv = VEnv.singleton id (typ, iters) in
+  add_vars_bind venv iterctx
+
+(* Filtering variables *)
+
+let filter_bound (f : Id.t -> typ -> iter list -> bool) (iterctx : t) : t =
```
jaehyun1ee: (Note) A note for myself, refactor it to `f : Var.t -> bool` (if it makes sense).

p4spec/lib/pass/elaborate/dimension.ml
```diff
@@ -7,35 +7,156 @@ open Error
...
+                 (Id.to_string id) (Typdim.to_string typ_min)
+                 (Typdim.to_string typ)))
+          (List.tl typs_sorted);
+        typ_min)
+      dctx
+end
+
+(* Expression inference *)
+
+let rec infer_exp (exp : exp) (iters : iter list) (dctx : Dimctx.t) : Dimctx.t =
```
jaehyun1ee: Minor, but why put `(dctx : Dimctx.t)` at the end? Can we reorder it to `let rec infer_exp (dctx : Dimctx.t) (exp : exp) (iters : iter list)`, and likewise for others?

p4spec/lib/pass/elaborate/dimension.ml
```diff
@@ -7,35 +7,156 @@ open Error
...
+  let typ', at = (exp.note, exp.at) in
```
jaehyun1ee: Minor, but I would put it `let typ = exp.note $ exp.at`, since L43 seems like the only use case.

p4spec/lib/pass/elaborate/dimension.ml
```diff
@@ -7,35 +7,156 @@ open Error
...
+
+let rec infer_prem (prem : prem) (iters : iter list) (dctx : Dimctx.t) :
+    Dimctx.t =
+  match prem.it with
+  | RulePr (_, notexp, _) -> infer_notexp notexp iters dctx
+  | IfPr exp -> infer_exp exp iters dctx
+  | IfHoldPr (_, notexp) -> infer_notexp notexp iters dctx
+  | IfNotHoldPr (_, notexp) -> infer_notexp notexp iters dctx
+  | LetPr _ ->
+      error prem.at "let premise should appear only after bind analysis"
```
jaehyun1ee: Maybe refine the message to: "only after the algo pass"? Likewise for other messages, if exists.

p4spec/lib/pass/elaborate/dimension.ml
```diff
@@ -255,131 +372,126 @@ and annotate_args (bounds : VEnv.t) (args : arg list) : VEnv.t * arg list =
...
-      (empty, occurs, prem)
-  | LetPr (exp_l, exp_r) ->
-      let occurs_l, exp_l = annotate_exp bounds exp_l in
-      let occurs_r, exp_r = annotate_exp bounds exp_r in
-      let prem = LetPr (exp_l, exp_r) $ at in
-      let occurs = union occurs_l occurs_r in
-      (occurs_l, occurs, prem)
+      (occurs, prem)
+  | LetPr _ ->
+      error prem.at "let premise should appear only after bind analysis"
```
jaehyun1ee: Applies for this message as well.

p4spec/lib/pass/algo/binding/multibind.ml
```diff
@@ -0,0 +1,173 @@
...
+  let sidecondition = IfPr exp $ id.at in
+  let iterctx =
+    let iters = iters @ Iterctx.iters_of iterctx in
+    let venv =
+      List.map (fun id -> (id, (typ, []))) (id :: id_rename :: ids_rename)
+      |> VEnv.of_list
+    in
+    List.map (fun iter -> (iter, [], [])) iters |> Iterctx.add_vars_bound venv
+  in
+  Iterctx.iterate_prem iterctx sidecondition
```
jaehyun1ee, line 58: Nice! Can you give me a high-level explanation of what is going on in L51: `let iters = iters @ Iterctx.iters_of iterctx`? More broadly, I want to understand more about `Iterctx.t`. Does it represent the "outer" iterators of a premise, given that a premise is wrapped with `IterPr`s?

p4spec/lib/pass/elaborate/dimension.ml
jaehyun1ee, line 1: Can you add a brief summary of what is going on in this file, at a high-level, as a top-level comment?

How is dimension info collected (in `Dctx.t`), what is its scope (premise? rule? clause? or an exp?), and how does annotation work?

jaehyun1ee (changes requested):
Thank you for the hard work, the code is neat :) I made some minor comments, where **(NOTE)** are for myself. The code looks good as-is, but I would like if some comment were added for comprehension and future maintenance.

p4spec/lib/pass/algo/binding/bind.ml
shyukahn: Restored the comment in a new commit.

p4spec/lib/pass/algo/binding/bind.ml
shyukahn: Restored the comment in a new commit.

p4spec/lib/pass/elaborate/dataflow/collectbind.ml
shyukahn: Restored the comment in a new commit.

p4spec/lib/pass/elaborate/elab.ml
shyukahn: Thanks for pointing it out! I revised `check_prems_internal` to only check whether multiple `otherwise` premises are used. As you have suggested, checking whether non-pure premises are used alongside an `otherwise` premise is moved over to the `algo` pass, during binding analysis.

p4spec/lib/pass/elaborate/dimension.ml
shyukahn: Reordered the parameters in a new commit.

p4spec/lib/pass/elaborate/dimension.ml
shyukahn: Applied the change in a new commit.

p4spec/lib/pass/elaborate/dimension.ml
shyukahn: Refined the error message in a new commit.

p4spec/lib/pass/elaborate/dimension.ml
shyukahn: Refined the error message in a new commit.

p4spec/lib/pass/algo/binding/dimension.ml
shyukahn: `p4spec/lib/pass/algo/binding/dimension.ml` is used to find free variables and their dimensions in expressions, and is a simplified (or perhaps duplicated…) version of dimension inference implemented in `p4spec/lib/pass/elaborate/dimension.ml`.

The `Typdim.sub` check is done to find the minimal dimension of the `id`, as an `id` may occur with different dimensions in an iterated premise. For example, in an iterated premise such as `-- (if ... = (n, n*))*` iterating over `n*`, inferring the RHS `(n, n*)` should result in `{ n : nat }`. This information is mainly used when reconstructing iterated premises during binding analysis.

p4spec/lib/pass/algo/binding/multibind.ml
shyukahn: In the RHS of L51(`iters @ Iterctx.iters_of iterctx`), `iters` is from the iterators from the expression that is renamed, and `iterctx` is from the outer iterators of the original premise.
```
-- (let (n*, n*) = ...)**
```
In the example above, multibinding analysis is implemented to rename the second occurrence of `n*` with a fresh `id`, say `n'*`. Then `iters` is `[ * ]` and `Iterctx.iters_of iterctx` is `[ *; * ]`. Since `n` here is a binding occurrence, its dimension can be determined by concatenating these two lists of iterators. The renamed `id`s have the same dimension. Therefore, after multibinding analysis, the resulting premise and sidecondition are:
```
-- (let (n*, n'*) = ...)**
-- (if n = n')***
```
I'll write about `Iterctx.t` in a separate comment in more detail.

p4spec/lib/pass/elaborate/dimension.ml
shyukahn: Added a brief explanation of dimension analysis with an example in the file in a new commit.

shyukahn:
Thank you for the review! I pushed some commits to address your feedback. Below is some detail on how `Iterctx.t` is intended to be used.

```OCaml
(* Iterctx *)
type t = (iter * var list * var list) list
```

As you mentioned, given a premise wrapped with `IterPr`s, `Iterctx.t` represents the outer iterators of the premise. Its scope is restricted to a single premise.

Binding analysis introduces renaming expressions and emits sidecondition premises. If the target is an iterated premise, we need to (1) remove/add renamed variables over which the premise iterates, and (2) generate corresponding iterated sidecondition premises. For (1), the iterators from nested `IterPr`s are converted into an `Iterctx.t`, which is threaded through binding analysis. For (2), an `Iterctx.t` is constructed manually for each kind of sidecondition, and `Iterctx.iterate_prem` is used to generate the final iterated side condition premise.

There are also some occasions in partial binding analysis where `Iterctx.t` is used to represent the iterators of an _expression_ instead of a premise.

```OCaml
module REnv = struct
  type t = (To.t * From.t * Iterctx.t) list
  ...
end
```

`REnv.t` stores an `Iterctx.t` for each renaming. When an expression wrapped with `IterE`s is renamed, the `Iterctx.t` representing the outer iterators of the expression is stored alongside it. This is later concatenated with the `Iterctx.t` of the premise's outer iterators when generating sideconditions, as in `gen_prems`:

```OCaml
let gen_prems (ctx : Ctx.t) (iterctx_prem : Iterctx.t) (renv : REnv.t) :
    prem list =
  List.concat_map
    (fun (to_, from_, iterctx_exp) ->
      gen_prem ctx to_ from_ (iterctx_exp @ iterctx_prem))
    renv
```
`Iterctx.t` values used for this purpose are explicitly named `iterctx_exp`. That said, there's room to improve this design, since the second `var list` of `Iterctx.t` is never used when it holds the iterators of an expression.

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1935,22 +1685,8 @@ and elab_tablerow (ctx : Ctx.t) (at : region) (id : id)
...
-      (fun arg_il ->
-        match arg_il.it with Il.ExpA exp_il -> exp_il | _ -> assert false)
-      args_il_signature
-  in
-  let _ctx_local, exp_il =
-    elab_tablerow_output_with_bind ctx_local typ_il exp_body
-  in
-  let tablerow_il =
-    (exps_il_signature, args_il_input, exp_il, sideconditions_il) $ tablerow.at
-  in
```
jaehyun1ee, line 1953: Now that this part of the code is gone (which checks shallow-binds for table arguments, or row patterns), the pattern-set computation operates on unchecked table rows.

```
and pattern_set_covered_by_arg (ctx : Ctx.t) (arg_il : Il.arg) :
    Pattern.PatternSet.t =
  let exp_il =
    match arg_il.it with Il.ExpA exp_il -> exp_il | Il.DefA _ -> assert false
  in
  match exp_il.it with
  | VarE _ -> pattern_set_covered_by_typ ctx (exp_il.note $ exp_il.at)
  | UpCastE (_, { it = VarE _; note; at }) ->
      pattern_set_covered_by_typ ctx (note $ at)
  | UpCastE (_, { it = CaseE notexp_il; at; _ }) ->
      let mixop, exps_il = Mixfix.split notexp_il in
      [
        Mixfix.fill mixop
          (List.map (fun exp_il -> exp_il.note $ exp_il.at) exps_il)
        $ at;
      ]
      |> Pattern.PatternSet.of_list
  | _ -> assert false
```

Here, if the table rows do *not* satisfy shallow bind, elaboration would end up in `assert false`, which is undesirable. Can we consider moving this part also over to the algo pass?

p4spec/lib/pass/algo/binding/dimension.ml
jaehyun1ee: Thanks for the explanation. Can you also add a high-level comment at the top of this file, likewise the `dimension.ml` in elab pass?

jaehyun1ee (changes requested):
Thanks! I have two outstanding comments, before I have a final pass and land this PR.

p4spec/lib/pass/elaborate/elab.ml
shyukahn: That's a good point. I migrated the pattern-set computation on tables rows after checking they are shallow binded, within the algo pass.

p4spec/lib/pass/algo/binding/dimension.ml
shyukahn: Added a top-level comment to `dimension.ml` in the algo pass in a new commit.

jaehyun1ee (approved):
Thanks for the hard work, merging!

PR #170, opened 2026-07-02
Static typing rules for nano-p4

pacokwon:
This PR adds typing rules for nano-p4.


nano-p4/spec/1-syntax.watsup
```diff
@@ -326,72 +248,61 @@ syntax assignmentStatement = lvalue `= expression `;
...
-syntax blockElementStatement =
-  | constantDeclaration
+syntax statement =
+  | emptyStatement
   | variableDeclaration
-  | statement
+  | assignmentStatement
+  | callStatement
+  | blockStatement
+  | conditionalStatement
```
jaehyun1ee, line 289: Applies overall, consider inlining `emptyStatement`, ..., `conditionalStatement` into `statement`, like:

```
syntax statement =
   | `;
   | ...
   | IF `( expression ) blockStatement ELSE blockStatement
```

This comes with pros and cons tho, the pro(s): shorter LoC for the syntax defn; the con(s): we can't use meta-variables like `conditionalStatement` no more, which sometimes come in handy.

nano-p4/spec/5.03-typing-expression.watsup
```diff
@@ -0,0 +1,100 @@
...
+    scope TC |- (`TID typeId) `( `EMPTY ) : parserObjectTypeIR
+    -- if PARSER parameterIR* = $find_callableDef_t(TC, typeId)
+    -- if parserObjectTypeIR = PARSER typeId `( parameterIR* )
+
+  rule Expr_ok/control:
+    scope TC |- (`TID typeId) `( `EMPTY ) : controlObjectTypeIR
+    -- if CONTROL parameterIR* = $find_callableDef_t(TC, typeId)
+    -- if controlObjectTypeIR = CONTROL typeId `( parameterIR* )
+
+}
```
jaehyun1ee, line 96: If so, why not restrict the `callExpression` grammar at the first place?

nano-p4/spec/5.01-typing-relation.watsup
```diff
@@ -0,0 +1,143 @@
...
+  typingContext |- type ~> typeIR
+  hint(input %0 %1)
+
+;;
+;; Type elaboration
+;;
+
+relation Type_alpha:
+  typeIR ~~ typeIR
+  hint(input %0 %1)
```
jaehyun1ee: Now that we don't have generics, maybe put it `Type_eq` instead?

nano-p4/spec/5.00-typing-context.watsup
```diff
@@ -0,0 +1,227 @@
...
+;;;; Finders for callableTypeDefs
+;;
+
+dec $find_callableDef_t(typingContext, callableId) : callableTypeDef
+def $find_callableDef_t(TC, callableId) = callableTypeDef
+  -- if callableTypeDef
+      = $find_map<callableId, callableTypeDef>(TC.GLOBAL.CALLABLE, callableId)
+
+;;
+;; typeDefIR -> typeIR. only called for monomorphic types
```
jaehyun1ee: Aren't all type monomorphic already?

jaehyun1ee (approved):
LGTM! Made minor comments. I think the current syntax defn is the nano-est we can get, but still, nano P4 is quite larger than what I initially imagined of. Let's later discuss how to document this, into an intuitive tutorial.

nano-p4/spec/5.01-typing-relation.watsup
pacokwon: Renamed `Type_alpha` to `Type_eq` and `ParameterType_alpha` to `ParameterType_eq`.

nano-p4/spec/5.00-typing-context.watsup
pacokwon: Outdated comment. The description has been removed.

nano-p4/spec/5.03-typing-expression.watsup
pacokwon: Good point. If you're referring to `argumentList` being empty, do you think
```
syntax callExpression = callTarget `( `EMPTY )
```
this is a good choice?

```
syntax callExpression = callTarget `(  )
```

doesn't seem to be allowed. not entirely sure why.

nano-p4/spec/1-syntax.watsup
pacokwon: I think we can leave it as is for now, but I will keep this in mind and consider what would work better when I work on documentation in the future.

pacokwon:
Merging for now, please revisit comments when you can!

Issue #38, opened 2025-12-18
Arrays of non-header types


jaehyun1ee:
This PR adds arrays of non-header types to the language, i.e., generalizes over the header stack syntax: https://github.com/p4lang/p4-spec/pull/1360. The mechanized spec should reflect this update.

PR #171, opened 2026-07-02
Specification of P4 arrays

shyukahn:
This PR specifies P4 arrays.

## P4 arrays

Arrays in P4 are a generalization of header stacks, which are arrays of headers.

`headerStackType` is renamed to `arrayType` in the P4 surface syntax.

```diff
- syntax headerStackType= namedType `[ expression ]
+ syntax arrayType = type `[ expression ]
```

`arrayValue` and `arrayTypeIR` are added for arrays that are not header stacks (i.e. arrays whose elements are not headers nor header unions). `headerStackTypeIR` is modified to distinguish header stacks with arrays that are not header stacks.

```diff
+ syntax arrayValue = ARRAY `[ value* ]
- syntax headerStackValue = HEADER_STACK `[ value* `( nat `; nat ) ]
+ syntax headerStackValue = HEADER_STACK `[ value* `( nat ) ]

+ syntax arrayTypeIR = ARRAY typeIR `[ nat ]
- syntax headerStackTypeIR = typeIR `[ nat ]
+ syntax headerStackTypeIR = HEADER_STACK typeIR `[ nat ]
```

## Test results

Among the 7 excluded P4 test cases in `excludes/static/future/merged/s-1360-arrays.exclude`, 5 tests now pass the static tests, including 1 P4 file with 2 STF files additionally passing the dynamic tests. 2 files are still excluded in `excludes/static/p4c-specific/implicit/array-with-declarator.exclude` due to the use of declarators when declaring an array. This can be seen as a p4c-specific feature, as it is not specified in the official P4 specification.

```P4
bit<8>[16] arr; // Supported
bit<8> arr[16]; // Not supported
```

1 negative test is additionally excluded in `excludes/static/p4c-specific/implicit/array-index-simplifier.exclude`. The failing cause is a limitation of a p4c pass.


jaehyun1ee:
LGTM overall, one comment regarding the type well-formedness check for type nesting rules.

`array`s can be nested within `struct`/`tuple`s, where `array`s permit some more types than `struct`/`tuple`s, e.g., `extern` types. Said again, `extern` can be nested within an `array`, but not in an `array` within a `struct`. Likewise the nested check for `header` types, please add such checks to the spec.

shyukahn:
Thanks for pointing it out! I've made another commit that checks `array` types nested within `struct`/`tuple`s separately.

jaehyun1ee (approved):
Thanks! Merging. Closes #38.

PR #173, opened 2026-07-05
Add @micro test, based on the p4c test suite

jaehyun1ee:
This is an alternative version of #166. It adds a `@micro`/`@micro-det` test suite, that runs fast, to be used for internal dev purposes. While the referenced PR uses some small, hand-written test corpus, this PR instead uses an `*.include` list to re-use parts of the p4c test suite as the test corpus.


KunJeong:
[@jaehyun1ee said:](https://github.com/kaist-plrg/p4-spectec/pull/166#issuecomment-4885525875)
> Thanks for the addition, a fast test suite was necessary for faster turn-around time. I expect that the corpus will grow over time, so given this expectation, seems like reusing parts of the official p4c test suite might be a better direction. I made a similar PR at #173, PTAL.

Carrying over from #166.

As we've discussed, the new tests are for faster loops in development. While I agree that delegating to p4c might be desirable if we need to expand the test suite continuously, `@micro` isn't really meant for long-term maintenance, and I believe we can freely remove it anytime we feel a maintenance burden.

1) A major point of #166 was not adding new OCaml test driver logic. Even though the `onlys` logic here is relatively trivial, adding a new CLI flag for a new test tier kind of defeats the purpose as now we're really testing a different execution path.

2) The three .p4 files in `pos.include` are good selections. But the 11 .p4 files in #166  have been hand-crafted to cover diverse types, casts, operations, and syntactic constructs including `enum`, `header_union`, and `switch` (all which are not in the three files), has 3547 (vs. 3198 current) spec instruction coverage via `Program_inst`, with a similar net runtime (0.83s vs 0.80s on my machine). Smaller programs mean better failure isolation and debugging as well; we both know the pain of debugging something like `action_synth.p4`.

So in terms of developer utility and trustworthiness, I believe #166 is preferable at the moment. Why don't we revisit the additional test driver logic / p4c delegation when we have an actual maintenance or scalability issue?

jaehyun1ee:
Closing in favor of #166

PR #177, opened 2026-07-16
Dynamic semantics for nano-p4

pacokwon:
This PR adds dynamic semantics for nano-p4.

Some design choices I made that might be worth giving more attention to:

## Absence of IR counterparts
The P4 spec written in P4-SpecTec elaborates the whole program into an IR form, that is then passed to the instantiation and execution phases.

The Nano-P4 spec takes a different approach. Once type checking is performed, types are (almost) erased. The only case where elaborated types are used is during initialization of variables.

For other instances where type information is needed during execution, the resulting type definition environment from type checking phase is inherited by the evaluation context.

## Instantiation
Since Nano-P4 doesn't have any stateful components, instantiation isn't necessary. However, there *is* a loading phase inserted in between type checking and execution. This phase:

* Loads the parser block and filter block used by the architecture. The execution phase uses this to run each block.
* Loads callables (actions and constructors) into the context. The execution phase uses this to find callable definitions.

As can be seen, this isn't an instantiation phase in terms of functionality, so its name is open to change.


jaehyun1ee (commented):
Sorry for the late review, seems like the recent update to the codebase is conflating the diffs in this PR. ~The diffs show that `nano-p4/spec` has been deleted. Could you kindly open a new PR s.t. it shows the actual additions to the nano-p4 dynamic semantics?~

My bad. The nano-p4 spec is a submodule now. I'll take a look there and comment separately.

pacokwon:
Yeah that was actually my bad. Thank you!

jaehyun1ee:
#### On Absence of IR counterparts

Sounds and looks good, one minor comment on the placement of "(almost) erased" types:

```
syntax argumentIR = expression `# typeIR
syntax parameterIR = direction typeIR nameIR
```

These now live in `2-domain.watsup`. `2-domain.watsup` seems like a mixture of the *actual* domain defs (`id`, ...), `typeIR` defns, and `argumentIR/parameterIR` defns. Maybe split them out into three, smaller files?

#### Instantiation

It is a good design, and I agree. This greatly reduces the complexity of the spec.

One question about the description:

> Since Nano-P4 doesn't have any stateful components, instantiation isn't necessary.

So, in Nano-P4, tables do *not* retain their state across packets? What happens to tables when we feed STF `add/remove` commands to add and remove table entries?

Also, given the design, we might want to just rename "instantiation" to "load", then clarify their differences in the tutorial docs.

pacokwon:
> These now live in 2-domain.watsup. 2-domain.watsup seems like a mixture of the actual domain defs (id, ...), typeIR defns, and argumentIR/parameterIR defns. Maybe split them out into three, smaller files?

This sounds good. I can do that.

---

> So, in Nano-P4, tables do not retain their state across packets? What happens to tables when we feed STF add/remove commands to add and remove table entries?

The current design of the language/architecture is that they do *not* support control-plane operations. So no support for `add`, `remove` directives in STF, just `packet` and `expect`. This is also reflected in the fact that a table either:

- Does *not* specify entries and (always) results in a no-op on `.apply()`, or
- Specifies entries but only through `const entries = { ... }`.

---

> Also, given the design, we might want to just rename "instantiation" to "load", then clarify their differences in the tutorial docs.

I agree. We can rename to "load".

pacokwon:
I made the mistake of extracting the spec to a submodule, so now it's hard to track changes here. But the discussed changes have been made to the [spec repo](https://github.com/pacokwon/nano-p4-spec)!

PR #184, opened 2026-07-22
Optimization for #163: Incremental hashing for compiled native OCaml types

jaehyun1ee:
## What

Every composite compiled OCaml type is now `(body, Il.vnote) note_phrase`, the same shape named types already use. Tuples, lists, options, and their nested/generic forms carry a note holding a fresh `vid`, the reified type, and an incremental `vhash`. This gives composites an O(1) structural hash (a slot read) and a `== / vid / vhash` short-circuit equality, closing the gap that forced the ML tower into deep structural comparison for tuples/lists/options.

## Codegen (`gen/ast/{exp,bind,instr}.ml`)

- Construction wraps through a shared smart constructor; destructuring reads `.it`.
- `Binding.t` now carries the SpecTec type, so iteration-list/option bindings, iteration expressions, and instruction iterated-lets/rules re-wrap the bare lists/options their `splitN`/`fold` produce:

```ocaml
type t = Var.t * Ml.id * Sl.typ
```

- `gen/wrap.ml` (new) holds `compile_mk_wrap` / `compile_wrap_hash`, shared by `exp` and `bind` without a `Bind -> Ast` module cycle. A composite with no monomorphic `mk` (nested or generic) wraps inline via `Converter.wrap_note` instead:

```ocaml
| Il.TupleT _ | Il.IterT _ ->
    Interface.Converter.wrap_note tparams typ_expand expr_body_ml
      (compile_wrap_hash ~tparams ctx typ_expand)
```

## Boundary (`backend-ocaml-{sl,p4,il}/val_native.ml`)

`Get` unwraps the note body (field 0); `Make` stamps a note. `child_hash` dispatches on representation — the `note_phrase` record is the only tag-0, size-3 block — so the boundary needs no spec type:

```ocaml
let child_hash (x : t) : int =
  if Obj.is_block x && Obj.tag x = 0 && Obj.size x = 3 then
    (Obj.obj x : (t, Il.vnote) note_phrase).note.vhash
  else Hashtbl.hash x
```

## Fixes surfaced by the ebpf packet sim

- `val_native.child_hash` no longer slot-reads a bare prim-alias (`bit<9>`, externs), which segfaulted on values that carry no note.
- `gen/interface/note.ml` seeds a variant case hash by `Hashtbl.hash ctor_ml` (the ctor name), not its positional index. With an index seed the same ctor (`` `DEFAULT ``) hashed differently in `simple_keyset_expression` vs `keyset_expression`, so the `vhash` eq short-circuit wrongly rejected two equal `` `DEFAULT `` values across a variant-inclusion coercion — which broke `default`/`_` select keysets (compared via `eq`). A type-independent seed keeps a value's `vhash` stable across variant subtyping, restoring soundness.

## Result

ebpf packet sim (ML/compiled mode): **2/17 → 15/17**. `Program_ok` and `Program_init` both pass in ML. The remaining two (`ebpf_checksum_extern`, `ebpf_conntrack_extern`) fail in `EBPF_filter` on the checksum/conntrack extern paths, unrelated to note-wrapping.

## Verify

```
make build && make gen-ocaml && make build-compiled
make gen-ocaml-il && make gen-ocaml-sl && make build-boot-comp
cd p4spec && opam exec -- dune build @sim-ebpf-p4c-ml --profile=release
```

### DIFF
--- a/p4spec/lib/backend-ocaml-il/val_native.ml
+++ b/p4spec/lib/backend-ocaml-il/val_native.ml
@@ -16,7 +16,7 @@ module Mixop = Domain.Mixop
 module Il = Lang.Il
 module Num = Lang.Xl.Num
 
-(* This library's own generated dispatch tables (spec-meta/il), not P4's. *)
+(* This library's own generated dispatch tables (spec-meta/sl), not P4's. *)
 module Spec_parts = Spec_parts_il
 open Util.Source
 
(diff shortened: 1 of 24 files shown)


jaehyun1ee:
Superseded by #190 (recreated with renamed branch after this PR's head branch was renamed). Continuing there.

PR #185, opened 2026-07-22
Compiled simulator: cross the extern boundary by cast, convert only to persist

jaehyun1ee:
## What this does

The compiled (`ML_mode`) simulator runs a P4 program as native OCaml. Every value it hands to an extern or a builtin — and every value handed back — used to be **deep-converted** between the generated OCaml representation and the generic self-describing `Value.t` tree, and converted back on return. For a program with heavy table/packet activity that is millions of `Value.t` nodes built and thrown away, and the conversion dominates the run.

That conversion is unnecessary on the common path: a value that leaves compiled code and is handed straight back to compiled code never needs to be inspected as a `Value.t`. It only has to become a real `Value.t` when it is actually **stored** somewhere serialized (a register, the scheduler's packet context).

This change separates the two cases behind one interface, `Valrep.VAL`. The compiled path passes values across the extern boundary as-is with an O(1) cast, and pays a real conversion only when a value is genuinely persisted. The tree-walking `IL_mode`/`SL_mode` interpreters are unchanged — they still use `Value.t` end to end, now as one instantiation of the same interface.

## How it works

`Valrep.VAL` describes what the extern, builtin, and simulator layers need from a value. It has two kinds of cross-boundary operation, and that split is the whole idea:

```ocaml
(* pass-through: value goes straight back to compiled code, never read here *)
val to_value : t -> Value.t
val of_value : Value.t -> t

(* convert: value is about to be stored in a serialized field, so it must
   become a real Value.t of its statically known spec type *)
```

Two implementations satisfy it:

- `V_value` — `t = Value.t`, used by the `IL_mode`/`SL_mode` interpreters. Every operation is the existing `Value.t` operation; all four crossings above are the identity.
- `V_native` — `t = Obj.t` holding the compiled spec's native OCaml value, used by `ML_mode`. `to_value`/`of_value` are `Obj.repr`/`Obj.obj` casts (the fast path); `marshal`/`unmarshal` do a real per-spec-type conversion (the only place the old cost is still paid, and only on a store).

## Example

A generated extern dispatcher, before — it deep-converts each argument and the result:

```ocaml
let a_i = unmarshal_<typ> (List.nth args__ i) in
...
```

and after — it casts in and casts out, so the extern runs directly on the native values:

```ocaml
let a_i = (Obj.magic (List.nth args__ i) : <ocaml typ>) in
...
```

The generated builtin and extern-relation wrappers change the same way, so the compiled interpreter never marshals across those boundaries either.

## The changes, in order

**1. The `Valrep` abstraction.** `runtime/valrep/` adds the module types `SAFE` (structural: `at`, `marshal`/`unmarshal`, `Get`/`Make`), `UNSAFE` (`to_value`/`of_value`, the unchecked reinterpret only the per-target boundary sees) and `VAL = SAFE + UNSAFE`, plus `V_value` — the identity instantiation over `Value.t`. Nothing consumes it yet.

**2. Generated `marshal_typed`/`unmarshal_typed`, keyed on `Typ.t`.** The per-spec-type conversion `V_native.marshal`/`unmarshal` dispatch through. Both are one `match` on the type's head name, with structural arms for `IterT` (lists, options) and `TupleT` so a bare `val*` marshals without a hand-written helper. The entry program type is threaded from the CLI (`"p4program"` for the P4 spec, `"script"` for the meta specs) rather than hardcoded, so the closure is seeded correctly for every spec.

**3. `V_native` and the generated case bridge.** `backend-ocaml-p4/val_native.ml` is the native representation: `t = Obj.t`, cast crossings, and `marshal`/`case`/`make_case` dispatching through the generated `Spec_parts_p4.Dispatch.*`. `case_of_typed`/`make_case_typed` read and build a native variant given its spec type. Container types (`set`, `pair`, `map`, `res` and their aliases) are handled by the **normal** variant path with their type parameters erased to `Obj.t` and aliases resolved from the spec — not a hardcoded list of head names. Every generated inner match ends in a wildcard that fails with a named error, so a value of the wrong type fails fast instead of projecting garbage.

**4. Cross the extern boundary by cast.** The generated `ML_mode` dispatchers (`eval_func_native`/`eval_rel_native`) and the generated builtin, extern-function and extern-relation wrappers all pass typed `Obj.t` through the `Value.t` wire as an `Obj.magic` cast, dropping the marshal/unmarshal converters. The interpreted `eval_func`/`eval_rel` and their converters are untouched; `marshal`/`unmarshal` survive only on the persist path.

**5. The simulator, generic over the representation.** `backend-sim` and its builtins become a functor over `V : Valrep.VAL`; `ARCH` gains a `vt`, `INTERFACE` gains a `Builtins` functor. The simulator then instantiates two stacks — one at `V_value` for `IL_mode`/`SL_mode`, one at `V_native` for `ML_mode` — and `init_mode` selects the active one. The parsed program is `unmarshal`ed to its native type exactly once at entry; the per-mode program-init trampoline returns native ctx/arch that the pipeline `of_value`-casts, since they are already native.

**6. `DebugI` prints under `ML_mode`.** `-- debug exp` compiled `ML_mode` used to raise a swallowed `Unmatch` and print nothing. It now prints the same source line and value the interpreted path does — a small tool that made the rest of this change debuggable.

## What is deliberately left as-is

- **`Value.t` is the extern wire type**, unconditionally, in the shared runner signature. Making it representation-parametric would cascade into parameterizing the `IL`/`SL` interpreters over `Valrep` too, which is far more than this change is worth. So every boundary keeps exactly one well-typed crossing, commented where it happens.
- **The three `V_native` files** (`backend-ocaml-{p4,il,sl}/val_native.ml`) are near-identical, differing only in which generated `Spec_parts_*` table they bind. They are kept as separate copies rather than one functor, matching the per-target library split, with a comment on each.

## Verification

- `make test-run-ml` — the compiled interpreter runs every P4 sample under `ML_mode` and matches the interpreted result (1307/1307).
- `make test-sim-v1model-p4c-ml` — the compiled v1model simulator runs the STF packet tests under `ML_mode` and matches the `SL_mode` output.
- `make test-fast` — the interpreted `IL`/`SL` paths are unchanged.

---

# Part 2 — the boot tower

## What this does

`spectec-boot-comp` runs a *tower* of specs: the P4 spec at the bottom, and above it `spec-meta/il` / `spec-meta/sl` — the formal semantics of SpecTec itself — each interpreting the level below. Those meta levels were always tree-walked. This part lets a meta level run as **compiled native OCaml** (`ML_mode`) too, reusing the same `Valrep.VAL` interface: the boot/unboot, interface, and boot-layer code is made generic over the representation, and each tower level picks the representation its mode needs.

## How it works

Four things became generic over `V`, and one routing changed:

- **boot/unboot** (`interface/spectec/{common,ili,sli}`) — the translation between an OCaml AST node (`Il.exp`, `Il.value`, …) and a value in whichever representation is running. Now `Make (V : Valrep.SAFE)` functors, threading `V` through construction (`V.Make`) and destruction (a generic `Dispatch` over the type's mixop).
- **the SpecTec interface** (`interface/interface.ml`) — split into a fixed boot-time part plus `Make (V : Valrep.VAL)` for the mode-dependent members.
- **the boot-layer externs** (`backend-boot/spectec.ml`) — `Make_null`/`Make_parametric` and `INTERFACE_SPECTEC` gain an abstract `vt`; the pairing constraint `INTERFACE_SPECTEC with type vt = V.t` turns a mispaired functor into a compile error instead of a silent representation-confusion bug.
- **the per-target native representation** — `backend-ocaml-il`/`backend-ocaml-sl` each get their own `V_native`, bound to their own generated `Spec_parts_*` tables.

`build.ml` then picks the representation once per level: `ML_mode` at a given interface selects that interface's `V_native`, everything else selects `V_value`.

## Example

A boot/unboot layer, before — hardcoded to `Value.t`:

```ocaml
let rec unboot_iter (value_iter : Value.t) : Il.iter = ...
```

after — generic, threading `V` and dispatching through the type witness:

```ocaml
module Make (V : Runtime.Valrep.SAFE) = struct
  module Dispatch = Runtime.Valrep.Dispatch.Make (V)
  let unboot_iter (value_iter : V.t) : Il.iter =
    Dispatch.dispatch value_iter Il_typs.typ_iter unboot_iter_tbl
      (fun _ _ -> error "@unboot_iter")
end
```

## The changes, in order

**1. Generic dispatch helper + type witnesses.** `Runtime.Valrep.Dispatch` splits a value's mixop shell (given its spec type) and runs the matching handler. `Il_typs` adds the `Il.typ` constants that key those dispatches.

**2. boot/unboot functorized over `V`** across the `common`, `ili`, and `sli` layers.

**3. Per-target `V_native` for il/sl**, each bound to its own `Spec_parts_*`; the `compiled_stub` gains the typed-bridge entries so the surface stays stable under `make build`.

**4. The SpecTec interface split** into a fixed part plus `Make (V)`.

**5. The boot-layer externs functorized** over `Valrep`, and `build.ml` picks the representation per (mode, interface).

**6. Per-level tower mode.** Each tower level carries its own mode instead of one tower-wide mode, so the boot level can run `ml` while the target runs `sl`.

**7. Builtin calls resolve through the interpreter.** A builtin call (`$add_map`, `$find_map`, a script's own `$print_*`) is routed through `Runner.Interp.eval_func`, which walks the tower level by level and re-enters itself at each hop — so a builtin owned two levels down is reached transitively, not by a single hop that only works for a two-level tower.

**8. The meta-cache under `ML_mode`.** It hashes a call's argument values as keys; a native argument tree must be marshalled to a real `Value.t` first (identity under `V_value`). Because that marshal costs more than recomputing, `init_mode` turns the meta-cache off under `ML_mode` and keeps it for the interpreted modes.

## Verification (boot tower)

The debug instruction (`-- debug`) prints under `ML_mode`, which is how the tower runs were checked. On `spectec-boot-comp`:

- `il-comp.json` (boot `spec-meta/il` compiled, P4 target interpreted) on `action-bind.p4` and `action-synth.p4` — both run to `passed`, matching the interpreted tower.
- `il-il-comp.json` (two compiled meta levels) on `action-bind.p4` — the builtin relay reaches two levels down and runs without fault.


jaehyun1ee:
Superseded by #191 (recreated with renamed branch after this PR's head branch was renamed). Continuing there.

PR #189, opened 2026-07-23
Bypassing the `Value.t` abstraction in builtin/extern calls from the compiled OCaml

jaehyun1ee:
## Why

When the compiled (ML) simulator calls an extern — or an extern calls back — every argument and result used to be fully converted between the generated OCaml representation and the generic `Value.t` tree. For the Dash v1model run that meant building roughly **27.7M `Value.t` nodes**, and that conversion was the dominant cost.

This conversion is unnecessary on that path: the value leaves the compiled code and is handed straight back to compiled code. It only has to become a real `Value.t` when it is actually **stored** somewhere that gets serialized (scheduler packet context, register payloads).

This PR separates those two cases. The compiled path passes values across the boundary as-is with an O(1) cast, and pays the conversion only when a value is genuinely persisted.

**Measured:** Dash v1model STF ~12.8s → ~4.3s wall; full `@sim-ml` suite roughly halved. The IL/SL interpreters are unchanged — they still use `Value.t` end to end.

## How

A small interface, `VAL`, describes everything the extern/builtin layers need from a value (destructors, constructors, case build/match). The extern, builtin, hash, packet, and state layers are made generic over it, so the same code runs against two implementations:

- **`V_value`** — `t = Value.t`. Used by the IL/SL interpreters. Every operation is the existing `Value.t` operation; the cross-boundary operations are the identity.
- **`V_typed`** — `t = Obj.t` holding the compiled spec's native OCaml value. Used by the compiled simulator. Crossing the boundary is `Obj.repr` / `Obj.obj`.

The interface has two distinct cross-boundary operations, which is the heart of the change:

```ocaml
(* Pass-through: value goes straight back to compiled code, never read here. *)
val to_value : t -> Value.t
val of_value : Value.t -> t

(* Convert: value is about to be stored in a serialized field, so it must
   become a real Value.t of its (statically known) spec type. *)
```

Under `V_value` both pairs are the identity. Under `V_typed`, `to_value`/`of_value` are casts (the fast path), and `marshal`/`unmarshal` do a real per-type conversion (the only place the old cost is still paid, and only when something is being stored).

## Concrete code changes

**1. Generated extern dispatch — convert → cast.** Inputs and outputs of compiled extern calls no longer go through `unmarshal_*` / `marshal_*`:

```ocaml
(* before: deep-convert each argument *)
let a_i = unmarshal_<typ> (List.nth args__ i)
...

(* after: cast in, cast out *)
let a_i = (Obj.magic (List.nth args__ i) : <ocaml typ>)
...
```

The extern wrappers in `rel.ml` / `func.ml` change the same way, which is what drops the two costly conversions (`marshal_eval_context`, `marshal_arch`) from the hot path.

**2. The program is converted exactly once.** `eval_program` receives a real parsed `Value.t`, so it `unmarshal`s it to a `p4program` a single time, then runs the typed path.

**3. Two parallel stacks, one chosen per run** (`make.ml`). A `V_value` stack for the IL/SL interpreters and a `V_typed` stack for the compiled simulator; `init_mode` wires up only the active one.

**4. Case build/match now take the value's spec type.** `Value.t` carries its own constructor tag at runtime, but a bare `Obj.t` is type-erased, so `V_typed` needs the spec type to pick the right OCaml projection. This also removed some hardcoded type-name guesses that previously mis-cast values.

---
_Recreated from #163 (auto-closed when its head branch was renamed `compiler-val` → `compiler-backend-native-val`)._


jaehyun1ee:
Closed, to be superseded by #191.

PR #249, opened 2026-07-25
Pin the opam switch in the make targets

KunJeong:
## Motivation

When a machine has multiple opam switches, the make targets ran under whichever switch the caller has active. The current `Makefile` doesn't handle this correctly: the `opam exec` calls names no switch, so it uses the caller's active switch instead. The `opam switch 5.1.0` line preceding each command mutates opam's global default as a side effect, but does not override the active switch that `opam exec` resolves against. Thus, two switches with different package versions can lead to local builds failing while CI succeeds or vice versa, or formatting shift when the active switch has a different version of `ocamlformat`.

## Core Concept
Opam resolves the switch in the following precedence:
explicit `--switch` > active switch > global default
The active switch is established by `eval $(opam env)`, and the global is set by `opam switch <name>`.

Each opam invocation is pinned with `opam exec --switch=5.1.0`, and the standalone `opam switch 5.1.0` lines are removed. Now, `make fmt` and `make build` behave the same regardless of the caller's active switch. This has the added advantage of leaving the global default alone, which causes confusion when working with multiple switches. 

## Commit Log

- chore(make): pin the opam switch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,24 +11,21 @@ EXEBOOT = _build/default/p4spec/bin/boot.exe
 build:
 	rm -f ./$(SPEC)
 	rm -f ./p4spec/lib/parsing/parser.ml ./p4spec/lib/parsing/parser.mli
-	opam switch 5.1.0
-	cd p4spec && opam exec -- dune build bin/main.exe && echo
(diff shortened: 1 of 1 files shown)


jaehyun1ee:
LGTM, thanks!

jaehyun1ee (approved):

Issue #251, opened 2026-07-26
typedef with an inline enum body makes the type unusable, so valid programs are rejected


jaehyun1ee:
A `typedef` whose body is an inline enum declaration, `typedef enum E { A } Alias;`, produces a type that cannot be used afterward. Neither the enum name `E` nor the alias `Alias` resolves, so any program that uses the type is rejected at type checking, even though the same enum written separately works.

This is a type-checking bug (a valid program is rejected), not a run-time bug.

## Minimal example

`bad.p4`:

```p4
#include <core.p4>

typedef enum E { A } Alias;

E f() { return E.A; }
```

## Steps

```bash
./p4spectec run spec -i p4c/p4include -p bad.p4 -rel Program_ok
```

Type checking fails at `E.A`:

```
relation Decl_ok failed
  ...
  relation Expr_ok failed        (on `E.A` in the body of f)
```

It should pass — the same enum declared separately does:

```p4
#include <core.p4>
enum E { A }
typedef E Alias;
E f() { return E.A; }        // passes
Alias g() { return Alias.A; } // passes
```

Every use of the inline form is rejected, not just `E.A`. All of the following fail type checking, while their separate-declaration equivalents pass:

```p4
typedef enum E { A } Alias;  E     f() { E x = E.A; return x; }   // fail
typedef enum E { A } Alias;  Alias f() { return Alias.A; }        // fail
typedef enum E { A } Alias;  E     f() { E x; return x; }         // fail  (E as a type)
typedef enum E { A } Alias;  Alias g() { Alias x; return x; }     // fail  (Alias as a type)
```

So `typedef enum E { A } Alias;` exposes neither `E` nor `Alias` as a usable type.

## Cause

The declaration is handled by `Decl_ok/derivedTypeDeclaration-mono` / `-poly` in `spec/5-typing/5.11-typing-declaration.watsup:998-1029`. The rule:

1. checks the inner declaration `enum E { A }`, producing a context `TC_1` that contains `E`;
2. builds the alias as `TC_2 = $add_typeDef_t(GLOBAL, TC_0, nameIR, typedefTypeIR)` — note this adds the alias to `TC_0`, which does **not** contain `E`;
3. but the rule concludes with `TC_1`, not `TC_2`.

So the alias in `TC_2` is discarded, and the returned `TC_1` does not carry a usable form of the type either. The result is a context in which neither name can be used.

[...]

Issue #245, opened 2026-07-24
Table size referring to a control variable type checks but fails to instantiate


jaehyun1ee:
A table property such as `size` may refer to a variable declared in the enclosing control, and this passes type checking. Building the program then fails, because the control's variables have no value yet when the table is constructed, so the property expression cannot be resolved.

## Minimal example

The table's `size` refers to the control-local variable `n` (`tsize.p4`):

```p4
#include <core.p4>
control C();
package Top(C c);
control MyC() {
    bit<8> n = 1;
    action a() { }
    table t {
        actions = { a; }
        default_action = a();
        size = n;
    }
    apply { t.apply(); }
}
Top(MyC()) main;
```

## Steps

```bash
./p4spectec run spec -i p4c/p4include -p tsize.p4 -rel Program_ok       # passes
./p4spectec run spec -i p4c/p4include -p tsize.p4 -rel Program_inst
```

Instantiation fails resolving `n`:

```
relation Expr_inst failed
└── spec/7-instantiation/7.04-inst-expression.watsup:36
    $find_var_i(prefixedNameIR, p, IC) failed
```

Replacing `size = n;` with `size = 1;` makes `Program_inst` pass, so the reference to `n` is the trigger.

## Cause

Control-local variables become visible to later declarations during type checking (`spec/5-typing/5.14.2-typing-control-declaration.watsup:25-29`), and a table's `size` (and other custom properties) need not be compile-time known (`spec/5-typing/5.14.1-typing-control-table.watsup:874-908`). So `size = n` type checks.

During control construction, a variable declaration produces no value binding: `ControlLocalDecl_inst/variableDeclarationIR` passes the declaration through unchanged (`spec/7-instantiation/7.10.2-inst-control-declaration.watsup:26-28`). But when the table is constructed, its custom properties are evaluated immediately: `TableProperty_inst/custom` runs `Expr_inst` on the `size` expression (`spec/7-instantiation/7.10.1-inst-control-table.watsup:132-139`). `Expr_inst` then looks up `n` with `$find_var_i` and fails because `n` was never bound (`spec/7-instantiation/7.02-inst-context.watsup:273-278`).

## Suggested fix

Bind a control-local variable's value at construction time, mirroring how a control-local constant is bound (`spec/7-instantiation/7.07-inst-declaration.watsup:11-14`), so later table properties can resolve it. Split the pass-through rule into an initialized and an uninitialized case.

```diff
[...]

Issue #243, opened 2026-07-24
Duplicate largest_priority_wins properties type check but get stuck at tie-breaking


jaehyun1ee:
A table may declare the `largest_priority_wins` property more than once and still type check. When two entries then tie and the tie-break consults this property, the run gets stuck: the helper that reads it only handles zero or exactly one occurrence, not two.

## Minimal example

`largest_priority_wins` is given twice, and two ternary entries both match (`duplpw.p4`):

```p4
#include <core.p4>
#include <v1model.p4>
header eth_t { bit<48> dst; bit<48> src; bit<16> ty; }
struct headers_t { eth_t eth; }
struct meta_t { }
parser prs(packet_in pkt, out headers_t hdr, inout meta_t meta, inout standard_metadata_t std) {
    state start { pkt.extract(hdr.eth); transition accept; }
}
control vfy(inout headers_t hdr, inout meta_t meta) { apply { } }
control ingress(inout headers_t hdr, inout meta_t meta, inout standard_metadata_t std) {
    action a() { std.egress_spec = 1; }
    action b() { std.egress_spec = 2; }
    table t {
        key = { hdr.eth.dst : ternary; }
        actions = { a; b; }
        default_action = a();
        const entries = {
            48w0 &&& 48w0 : a();
            48w0 &&& 48w0 : b();
        }
        largest_priority_wins = true;
        largest_priority_wins = false;
    }
    apply { t.apply(); }
}
control egress(inout headers_t hdr, inout meta_t meta, inout standard_metadata_t std) { apply { } }
control cmp(inout headers_t hdr, inout meta_t meta) { apply { } }
control dep(packet_out pkt, in headers_t hdr) { apply { pkt.emit(hdr.eth); } }
V1Switch(prs(), vfy(), ingress(), egress(), cmp(), dep()) main;
```

STF (`duplpw.stf`) — both ternary entries (`48w0 &&& 48w0`) match any key:

```
packet 0 000000000001 0000000000FF 0800
expect 1 000000000001 0000000000FF 0800
```

## Steps

```bash
./p4spectec run spec -i p4c/p4include -p duplpw.p4 -rel Program_ok       # passes
./p4spectec sim spec -arch v1model -i p4c/p4include -p duplpw.p4 -stf duplpw.stf
```

The simulation gets stuck applying the table:

```
runtime error: relation V1Model_ingress failed
└── spec/9-arch/9.A-v1model.watsup:402
    relation Stmt_eval failed
```

[...]

Issue #240, opened 2026-07-24
Parser state-local variables leak into the next state at run time


jaehyun1ee:
A variable declared inside one parser state is still visible in the next state at run time, even though the type checker treats each state as its own scope. The leaked variable shadows an outer one of a different width, so a value of the wrong width is written into a field and the deparser emits the wrong number of bytes.

## Minimal example

The parser declares `bit<16> x = 16w1` at parser scope, then a state-local `bit<8> x = 8w2` inside `start`. State `next` reads `x` (`leak.p4`):

```p4
#include <core.p4>
#include <v1model.p4>
header eth_t { bit<48> dst; bit<48> src; bit<16> ty; }
struct headers_t { eth_t eth; }
struct meta_t { }
parser prs(packet_in pkt, out headers_t hdr, inout meta_t meta, inout standard_metadata_t std) {
    bit<16> x = 16w1;
    state start {
        pkt.extract(hdr.eth);
        bit<8> x = 8w2;
        transition next;
    }
    state next {
        hdr.eth.ty = x;
        transition accept;
    }
}
control vfy(inout headers_t hdr, inout meta_t meta) { apply { } }
control ingress(inout headers_t hdr, inout meta_t meta, inout standard_metadata_t std) {
    apply { std.egress_spec = 1; }
}
control egress(inout headers_t hdr, inout meta_t meta, inout standard_metadata_t std) { apply { } }
control cmp(inout headers_t hdr, inout meta_t meta) { apply { } }
control dep(packet_out pkt, in headers_t hdr) { apply { pkt.emit(hdr.eth); } }
V1Switch(prs(), vfy(), ingress(), egress(), cmp(), dep()) main;
```

STF (`leak.stf`) — the `ty` field starts as `0800`:

```
packet 0 000000000001 0000000000FF 0800
expect 1 000000000001 0000000000FF 0800
```

## Steps

```bash
./p4spectec run spec -i p4c/p4include -p leak.p4 -rel Program_ok          # passes
./p4spectec sim spec -arch v1model -i p4c/p4include -p leak.p4 -stf leak.stf
```

The simulation runs to completion but emits the wrong packet:

```
runtime error: expected (1) 0000000000010000000000FF0800 but got (1) 0000000000010000000000FF02
```

[...]

PR #253, opened 2026-07-28
Patch relatively simple soundness bugs in the P4 spec (29 issues)

### LINKED ISSUE #197: Compound assignment in a parser does not propagate a reject from its right-hand side
In a parser, `x += e;` type checks, but if evaluating `e` produces a parser reject, there is no rule to continue and the interpreter gets stuck. Writing the same thing as `x = e;` works, so only the compound form is affected.

## Minimal example

`hs` is an empty header stack, so `hs.last` rejects with a stack-out-of-bounds (`ca.p4`):

```p4
#include <core.p4>
#include <v1model.p4>
header eth_t { bit<48> dst; bit<48> src; bit<16> ty; }
header H { bit<8> f; }
struct headers_t { eth_t eth; H[1] hs; }
struct meta_t { }
parser prs(packet_in pkt, out headers_t hdr, inout meta_t meta, inout standard_metadata_t std) {
    bit<8> x = 0;
    state start {
        pkt.extract(hdr.eth);
        x += hdr.hs.last.f;      // right-hand side rejects; compound assignment cannot continue
        transition accept;
    }
}
control vfy(inout headers_t hdr, inout meta_t meta) { apply { } }
control ingress(inout headers_t hdr, inout meta_t meta, inout standard_metadata_t std) {
    apply { std.egress_spec = 1; }
}
control egress(inout headers_t hdr, inout meta_t meta, inout standard_metadata_t std) { apply { } }
control cmp(inout headers_t hdr, inout meta_t meta) { apply { } }
control dep(packet_out pkt, in headers_t hdr) { apply { pkt.emit(hdr.eth); } }
V1Switch(prs(), vfy(), ingress(), egress(), cmp(), dep()) main;
```

STF (`ca.stf`):

```
packet 0 000000000001 0000000000FF 0800
expect 1 000000000001 0000000000FF 0800
```

## Steps

```bash
./p4spectec run spec -i p4c/p4include -p ca.p4 -rel Program_ok         # passes
./p4spectec sim spec -arch v1model -i p4c/p4include -p ca.p4 -stf ca.stf
```

The run fails with `relation Stmt_eval failed` on the `x += hdr.hs.last.f;` statement. Changing `+=` to `=` (same right-hand side) runs to completion.

## Cause

In `spec/8-dynamic/8.06-eval-statement.watsup`, the simple-assignment abort rule accepts any abort result from the right-hand side, but the compound-assignment abort rule only handles the exit case:

- simple assignment (`Stmt_eval/typedLvalueIR-cont-eq-typedExpressionIR-abort`, lines 25-37): right-hand side may yield any `abortResult`.
- compound assignment (`Stmt_eval/typedLvalueIR-cont-compound-typedExpressionIR-exit`, lines 39-51): right-hand side must yield `EXIT`.

A reject result from the right-hand side matches neither the compound-assignment abort rule nor the compound-assignment continue rule, so evaluation is stuck. A reject can reach here because `hs.last` on an empty stack rejects (`spec/8-dynamic/8.03.1-eval-expression.watsup:339-351`).

## Suggested fix

Let the compound-assignment abort rule accept any abort result, matching simple assignment:

```diff
--- a/spec/8-dynamic/8.06-eval-statement.watsup
+++ b/spec/8-dynamic/8.06-eval-statement.watsup
-  rule Stmt_eval/typedLvalueIR-cont-compound-typedExpressionIR-exit:
+  rule Stmt_eval/typedLvalueIR-cont-compound-typedExpressionIR-abort:
     p EC_0 ARCH_0 |- typedLvalueIR assignop typedExpressionIR `;
-                   : EC_2 ARCH_2 EXIT
+                   : EC_2 ARCH_2 abortResult
     ---- ;; evaluate l-value
     -- Lvalue_eval:
         p EC_0 ARCH_0 |- typedLvalueIR
                        : EC_1 ARCH_1 (`` storageReference)
     ---- ;; if compound assignment
     -- if assignop =/= `=
     ---- ;; evaluate expression
     -- Expr_eval:
         p EC_1 ARCH_1 |- typedExpressionIR
-                       : EC_2 ARCH_2 EXIT
+                       : EC_2 ARCH_2 abortResult
```

jaehyun1ee:
Fixes 29 soundness bugs where a P4 program type-checks but the spec's dynamic semantics get stuck or produce a wrong value — the `FIX-MECH ∧ EASY` slice of the soundness audit. One commit per bug, each carrying the spec fix plus a regression witness. Fixes group into four mechanisms: **MR** add a missing runtime rule, **WV** correct a wrong value, **IL** fix instantiation logic, **TT** reject the ill-formed program at typing.

`#229`/`#230` are fixed by the TT route (reject parameter aliasing) rather than the runtime route the issues proposed. `#251` is fixed in two parts — a typing-completeness fix and a parser fix registering a nested-`typedef`'s inner type name. `#225`/`#226` (size builtins) and `#235` (for-in element cast) are deferred out of this PR per review; `#225`/`#226` are tracked in #255.

The branch merges `concrete`, so every spec snippet below uses the per-kind atom markers (`` `< … `> ``, `';'`, `_NAME`, …) rather than the old backtick escape.

## Patches

| # | Class | Phase | Key change |
|---|---|---|---|
| [#198](https://github.com/kaist-plrg/p4-spectec/issues/198) | MR | dynamic | signed clause for `$un_bnot` |
| [#202](https://github.com/kaist-plrg/p4-spectec/issues/202) | MR | dynamic | signed clauses for `$bitacc_*_replace_op` |
| [#208](https://github.com/kaist-plrg/p4-spectec/issues/208) | MR | dynamic | string→string cast is identity |
| [#205](https://github.com/kaist-plrg/p4-spectec/issues/205) | MR | dynamic | `...` default-fill for aggregate casts |
| [#211](https://github.com/kaist-plrg/p4-spectec/issues/211) | MR | dynamic | sequence→sequence cast |
| [#201](https://github.com/kaist-plrg/p4-spectec/issues/201) | MR | dynamic | `.last` l-value rules for header stacks |
| [#237](https://github.com/kaist-plrg/p4-spectec/issues/237) | MR | dynamic | trailing fall-through switch label ⇒ eps |
| [#238](https://github.com/kaist-plrg/p4-spectec/issues/238) [#197](https://github.com/kaist-plrg/p4-spectec/issues/197) | MR | dynamic | propagate reject from parser if-else / compound-assign |
| [#240](https://github.com/kaist-plrg/p4-spectec/issues/240) | MR | dynamic | fresh local frame per parser state |
| [#219](https://github.com/kaist-plrg/p4-spectec/issues/219) | MR | dynamic | substitute type args for a generic local |
| [#234](https://github.com/kaist-plrg/p4-spectec/issues/234) | MR | dynamic | run an extern method from an object initializer |
| [#199](https://github.com/kaist-plrg/p4-spectec/issues/199) | WV | dynamic | header-stack `.size` is `bit<32>` |
| [#206](https://github.com/kaist-plrg/p4-spectec/issues/206) | WV | dynamic | label reordered struct-literal casts by target field |
| [#212](https://github.com/kaist-plrg/p4-spectec/issues/212) [#213](https://github.com/kaist-plrg/p4-spectec/issues/213) | IL | instantiation | offset-slice `+:` inst rule + ctk join |
| [#231](https://github.com/kaist-plrg/p4-spectec/issues/231) | IL | instantiation | thread explicit copy-in store into default copy-in |
| [#232](https://github.com/kaist-plrg/p4-spectec/issues/232) | IL | instantiation | name-based extern init, skip an omitted `@optional` param |
| [#245](https://github.com/kaist-plrg/p4-spectec/issues/245) | TT | typing | reject a non-compile-time-known table `size` |
| [#251](https://github.com/kaist-plrg/p4-spectec/issues/251) | IL | typing | keep both names of a type alias bound |
| [#251](https://github.com/kaist-plrg/p4-spectec/issues/251) | — | parse | register a nested `typedef`'s inner type name |
| [#204](https://github.com/kaist-plrg/p4-spectec/issues/204) | TT | typing | reject a scalar field in a header union |
| [#207](https://github.com/kaist-plrg/p4-spectec/issues/207) | TT | typing | reject duplicate struct-literal field names |
| [#222](https://github.com/kaist-plrg/p4-spectec/issues/222) | TT | typing | reject an undeclared type argument |
| [#227](https://github.com/kaist-plrg/p4-spectec/issues/227) | TT | typing | type `push_front`/`pop_front` as `VOID` |
| [#233](https://github.com/kaist-plrg/p4-spectec/issues/233) | TT | typing | reject an unimplemented abstract method |
| [#236](https://github.com/kaist-plrg/p4-spectec/issues/236) | TT | typing | loop body cannot guarantee a return |
| [#243](https://github.com/kaist-plrg/p4-spectec/issues/243) | TT | typing | reject duplicate `largest_priority_wins` |
| [#229](https://github.com/kaist-plrg/p4-spectec/issues/229) [#230](https://github.com/kaist-plrg/p4-spectec/issues/230) | TT | typing | disallow parameter aliasing |

## MR — add missing runtime rules

**#198** — `~` on a signed value had no `$un_bnot` clause:
```diff
+def $un_bnot(w S i) = w S i''
+  -- if i' = $bneg(i)
+  -- if i'' = $ite<int>($(i' >= 0), i', $($pow2(w) + i'))
```

**#202** — signed bit-slice assignment had no replace clause:
```diff
+def $bitacc_range_replace_op(integerValue_b, integerValue_h, integerValue_l, integerValue_r) = w S i
+  -- if w S i_b = integerValue_b
+def $bitacc_offset_replace_op(integerValue_b, integerValue_l, integerValue_w, integerValue_r) = w S i
+  -- if w S i_b = integerValue_b
```

**#208** — casting a string to string had no case:
```diff
+def $cast_string(STRING, stringValue) = stringValue
+def $cast_op(typeIR, stringValue) = $cast_string(typeIR_unroll, stringValue)
+  -- if typeIR_unroll = $unroll_typeIR(typeIR)
```

**#205** — aggregate cast with `...` default-fill (added for TUPLE/ARRAY/HEADER_STACK/STRUCT/HEADER and `$cast_record`; LIST excluded, its default is undefined in P4 7.3):
```diff
+def $cast_sequence(TUPLE `< typeIR* `>, SEQ `( value* ',' '...' `)) = TUPLE `( value_cast* `)
+  -- if (typeIR_given*, typeIR_rest*) = $partition_<typeIR>(typeIR*, |value*|)
+  -- if (value_given = $cast_op(typeIR_given, value))*
+  -- if (value_rest = $default(typeIR_rest))*
+  -- if value_cast* = value_given* ++ value_rest*
```

**#211** — sequence→sequence conversion had no case:
```diff
+def $cast_sequence(SEQ `< typeIR* `>, SEQ `( value* `)) = SEQ `( $cast_op(typeIR, value)* `)
```

**#201** — writing `hs.last` had no l-value rule (added eval/read/write mirroring `.next`, index `n_idx - 1`):
```diff
+  rule Lvalue_write/stack-last:
+    -- if n_idx_last = $(n_idx - 1)
+    -- if value_element_update* = value_element* [ [n_idx_last] = value ]
+    -- if headerStackValue_base_update = HEADER_STACK `[ value_element_update* `( n_idx `) `]
```

**#237** — a matching trailing fall-through switch label got stuck:
```diff
+  rule SwitchCases_table_eval/cons-fallthrough-match-no-block:
+    id |- switchCaseIR_h :: switchCaseIR_t* : eps
+    -- if switchLabelIR_h ':' = switchCaseIR_h
+    -- if $match_switchLabelIR_table(id, switchLabelIR_h)
+    -- if eps = $filter_<switchCaseIR>(switchCaseIR_t*, $is_non_fallthrough_switchCaseIR(switchCaseIR_t)*)
```

**#238 / #197** — a reject from a parser if-else condition, and an abort from a compound-assignment RHS, were not propagated:
```diff
-  rule Stmt_eval/typedLvalueIR-cont-compound-typedExpressionIR-exit:
-                   : EC_2 ARCH_2 EXIT
+  rule Stmt_eval/typedLvalueIR-cont-compound-typedExpressionIR-abort:
+                   : EC_2 ARCH_2 abortResult
+  rule ParserStmt_eval/else-reject:
+    -- Expr_eval: LOCAL EC ARCH |- typedExpressionIR_cond : EC_cond ARCH_cond rejectTransitionResult
```

**#240** — parser state-local variables leaked into the next state (wrong emitted bytes). Enter/exit a fresh local frame per state:
```diff
+    -- if EC_1 = $enter_e(EC_0)
     ...
+    -- if EC_3 = $exit_e(EC_2)
```

**#219** — an uninitialized generic local kept its unsubstituted type parameter; substitute through the LOCAL layer that carries the call's type args:
```diff
-    -- if typeIR_subst = $subst_type_e(BLOCK, EC_0, typeIR)
+    -- if typeIR_subst = $subst_type_e(LOCAL, EC_0, typeIR)
```

**#234** — an extern method bodied by an object initializer could not run. Added the callee rule that evaluates the stored body, plus a copy-in-abort companion so an aborting argument cannot re-stick:
```diff
+  rule Call_eval/copyin-cont-initializer:
+    -- if EXTERN_METHOD ... `{ theta ';' frame ';' blockStatementIR `} = externMethodCallee
+    -- DefinedFunctionBody_eval: EC_callee_5 ARCH_1 |- blockStatementIR : EC_callee_6 ARCH_2 definedFunctionResult
+  rule Call_eval/copyin-abort-initializer:
+    -- Copy_in: ... ~> ARCH_1 EC_1 EC_callee_4 '#' abortResult
```

## WV — correct wrong values

**#199** — header-stack `.size` yielded an infinite-precision value instead of `bit<32>`, in all three evaluators:
```diff
-    -- if expressionResult = _CONT (D n_size)
+    -- if expressionResult = _CONT (32 W n_size)
```

**#206** — an out-of-order struct literal labeled cast values with the source field name, landing them on the wrong fields:
```diff
-  = STRUCT typeId `{ (value_field_cast nameIR_field ';')* `}
+  = STRUCT typeId `{ (value_field_cast id_t_field ';')* `}
```

## IL — instantiation logic

**#212 / #213** — the offset-slice inst rule used range syntax, and the slice's ctk dropped the offset's dynamic status:
```diff
-    -- if expressionNoteIR = `( typeIR ctk_base_reduced `)
+    -- if ctk = $joins_ctk([ ctk_base_reduced, ctk_l_reduced, ctk_w_reduced ])
-    p IC STO_0 |- (typedExpressionIR_base `[ typedExpressionIR_l ':' typedExpressionIR_w `]) '#' _
+    p IC STO_0 |- (typedExpressionIR_base `[ typedExpressionIR_l '+:' typedExpressionIR_w `]) '#' _
```

**#231** — an inline constructor-argument allocation was rolled back, leaving a dangling store reference. Thread the explicit copy-in store into default copy-in (three constructor kinds):
```diff
-      STO_0 p IC
+      STO_1 p IC
```

**#232** — an omitted `@optional` extern constructor parameter was fetched unconditionally, and the native extern init read its arguments positionally, so a named/reordered call (`counter(type = …, size = …)`) crashed. `$init_objectState` now carries the parameter ids and the native inits resolve arguments by name:
```diff
-extern dec $init_objectState(nameIR, typeIR*, value*) : objectState
+extern dec $init_objectState(nameIR, typeIR*, id*, value*) : objectState
-  -- if objectState = $init_objectState(nameIR, typeArgumentIR*, value_cparam*)
+  -- if id_cparam* = id_aligned* ++ id_default*
+  -- if (value_cparam = $find_map<id, value>(IC_callee_4.BLOCK.FRAME, id_cparam))*
+  -- if objectState = $init_objectState(nameIR, typeArgumentIR*, id_cparam*, value_cparam*)
```

**#251** — adding a type alias dropped one of the two bound names; thread and return the updated context:
```diff
-    -- if TC_2 = $add_typeDef_t(GLOBAL, TC_0, nameIR, typedefTypeIR)
+    -- if TC_2 = $add_typeDef_t(GLOBAL, TC_1, nameIR, typedefTypeIR)
-          : TC_1 typedefDeclarationIR
+          : TC_2 typedefDeclarationIR
```
A `typedef enum E { A } Alias;` also registered only `Alias`, never the inner `E`, so `E.A` failed to parse. Register the inner name in the `typedef` rule's derived-type arm:
```diff
 typedef:
 	| t = typeRef
+		{ t }
 	| t = derivedTypeDeclaration
-		{ t }
+		{ declare_type (id_of_declaration t) (has_type_params_declaration t); t }
```

## TT — tighten typing (reject the ill-formed program)

**#245** — a table `size` bound to a non-compile-time-known value (a mutable control-local variable) type-checked, then failed to instantiate. Reject it at typing (P4 14.2.1.5); a literal or `const` size still passes:
```diff
   -- if (typeIR_unroll <: integerTypeIR)
         /\ ~(typeIR_unroll <: varBitTypeIR)
+  -- if DYN =/= $ctk_of_typedExpressionIR(typedExpressionIR)
```

**#204** — a header union admitted a scalar field. Sound to reject wholesale because P4 restricts `type` to scalar bases, so a newtype can never wrap a header:
```diff
 tbl dec $nestable_headerUnion(typeIR) : bool
-  | newTypeIR    => true
   | headerTypeIR => true
   | _            => false
```

**#207** — duplicate struct-literal field names passed typing:
```diff
+    -- if $distinct_<nameIR>(nameIR_f*)
```

**#222** — a bare-name type argument naming no declared type was accepted (in-scope type variables still resolve via `$bound`):
```diff
+    -- Type_wf: $bound(p, TC) |- _NAME typeId
```

**#227** — `push_front`/`pop_front` were typed as returning `int` but return no value; using the result now fails typing:
```diff
-  -- if methodTypeIR = BUILTIN_METHOD nameIR `( parameterIR `) ':' INT
+  -- if methodTypeIR = BUILTIN_METHOD nameIR `( parameterIR `) ':' VOID
```

**#233** — a named extern construction did not reject an object with an unimplemented abstract method:
```diff
+  -- if $is_concrete_extern_object(typeIR_object)
```

**#236** — a loop body was treated as always-executing in return-on-all-paths analysis; a zero-iteration loop cannot guarantee a return:
```diff
+  -- if TC_5' = $set_control_t(TC_5, TC_2.CONTROL)
-  -- if TC_6 = $set_loop_t(TC_5, TC_3.LOOP)
+  -- if TC_6 = $set_loop_t(TC_5', TC_3.LOOP)
```

**#243** — duplicate `largest_priority_wins` properties passed typing:
```diff
+def $is_tableLargestPriorityWinsProperty(_ _ tableCustomName _ ';') = true
+  -- if "largest_priority_wins" = $tableCustomName(tableCustomName)
+    -- if $(|$filter_<tableProperty>(tableProperty*, $is_tableLargestPriorityWinsProperty(tableProperty)*)| <= 1)
```

**#229 / #230** — a parameter default referencing a sibling parameter was accepted. `Parameter_ok` and `Parameters_ok` are now pure checks that never touch the context; `ParameterList_ok` folds the checked parameters into the context once, via a new `$add_parameters_t`. So every parameter — and every default — is checked under the pre-list context where no sibling is bound: a default naming a literal / `const` / enum member still passes, sibling aliasing is rejected, and the fold's `$add_var_t` still rejects a duplicate name:
```diff
 relation Parameters_ok:
-  cursor typingContext |- parameter* : typingContext parameterListIR '#' typeId*   ;; per-parameter update
+  cursor typingContext |- parameter* : parameterListIR '#' typeId*                 ;; pure check
 rule ParameterList_ok:
+  -- if TC_1 = $add_parameters_t(p, TC_0, parameterIR*)                            ;; single update
```

## Test harness

The regression corpus lives under `testdata/regression/{neg,pos,sim}`, one witness per issue, wired into the `run`/`sim` dune harnesses (`b879469f`), into `@micro` (`c6c6c0c1`), and into the `-det` variants (`d016ec86`) so `make test-micro` gates both correctness and determinism per commit:
```diff
+(rule
+ (target run_regression_neg_sl_det.actual)
+    (run ./test.exe run -sl -det -s ../../../spec -rel Program_ok
+     -i ../../../p4c/p4include -p4-dir ../../../testdata/regression/neg -neg)))))
```

## Regression corpus

```
neg/ (Program_ok -neg, must reject)   11 issues → 12/12 reject
pos/ (Program_inst, must pass)         3 issues →  4/4 pass
sim/ (sim -arch v1model)              17 witness → 17/17 pass
```

## Verification

```
make test-fast                                     # run+sim SL over the full p4c/p4testgen corpus — OK
make test-all-det                                  # -det over the full corpus — OK
dune build @run-sl @run-sl-det @sim-sl @sim-sl-det # OK
```

No p4c/p4testgen corpus program is newly rejected by the TT tightenings (`test-fast` green), and no added/overlapping rule introduces nondeterminism (`test-all-det` and the `-det` alias builds green). Every fix witness was cross-checked `-sl` against `-al`.

**Not run:** `make test-all`'s boot towers and the AL/PL interpreter passes.

Fixes #197 #198 #199 #201 #202 #204 #205 #206 #207 #208 #211 #212 #213 #219 #222 #227 #229 #230 #231 #232 #233 #234 #236 #237 #238 #240 #243 #251.


p4spec/lib/interface/p4/parser.mly
```diff
@@ -1299,8 +1299,11 @@ derivedTypeDeclaration:
 (* >> Typedef and newtype declarations *)
 typedef:
 	| t = typeRef
-	| t = derivedTypeDeclaration
 		{ t }
+	| t = derivedTypeDeclaration
+		{ (* a derived type nested in a typedef also introduces its own name *)
```
jaehyun1ee: Delete this comment.

p4spec/test/micro/programs/typedef_derived.p4
jaehyun1ee, line 1: Delete this file.

p4spec/test/micro/dune
```diff
@@ -925,10 +932,27 @@
    (diff micro_sim_ebpf_pl.expected micro_sim_ebpf_pl.actual)
    (diff micro_sim_psa_pl.expected micro_sim_psa_pl.actual)
    (diff micro_boot_al.expected micro_boot_al.actual)
-   (diff micro_boot_al_al.expected micro_boot_al_al.actual))))
+   (diff micro_boot_al_al.expected micro_boot_al_al.actual)
+   ;; Regression corpus: reuse the run-sl/sim-sl targets (testdata/regression)
```
jaehyun1ee: Delete this comment.

spec/5-typing/5.10-typing-statement.watsup
```diff
@@ -430,7 +430,9 @@ rule Stmt_ok/forStatement-three-clause:
       LOCAL TC_3 |- statement : TC_4 statementIR
   ---- ;; exit scope
   -- if TC_5 = $exit_t(TC_4)
-  -- if TC_6 = $set_loop_t(TC_5, TC_3.LOOP)
+  ---- ;; a loop body may run zero times; it cannot guarantee a return
+  -- if TC_5' = $set_control_t(TC_5, TC_2.CONTROL)
```
jaehyun1ee: No, give it a name `TC_6` and rename all following `TC_N`s to `TC_(N+1)`. I don't want `'`s.

spec/5-typing/5.10-typing-statement.watsup
```diff
@@ -430,7 +430,9 @@ rule Stmt_ok/forStatement-three-clause:
       LOCAL TC_3 |- statement : TC_4 statementIR
   ---- ;; exit scope
   -- if TC_5 = $exit_t(TC_4)
-  -- if TC_6 = $set_loop_t(TC_5, TC_3.LOOP)
+  ---- ;; a loop body may run zero times; it cannot guarantee a return
```
jaehyun1ee: Delete the comment.

spec/5-typing/5.10-typing-statement.watsup
```diff
@@ -534,7 +536,9 @@ rulegroup Stmt_ok/forStatement-in {
     -- Stmt_ok: LOCAL TC_3 |- statement : TC_4 statementIR
     ---- ;; exit scope
     -- if TC_5 = $exit_t(TC_4)
-    -- if TC_6 = $set_loop_t(TC_5, TC_3.LOOP)
+    ---- ;; a loop body may run zero times; it cannot guarantee a return
+    -- if TC_5' = $set_control_t(TC_5, TC_2.CONTROL)
```
jaehyun1ee: Also applies here.

spec/5-typing/5.11-typing-declaration.watsup
```diff
@@ -118,6 +118,8 @@ rule InstDecl_non_objectInitializer_ok:
       p TC_0 NAMED |- constructorTypeIR `< typeArgumentIR_infer* `# typeId_infer* >
                         `( argumentIR* `# id_default* `# id_optional* )
                     : typeIR_object `< typeArgumentIR_inferred* > `( argumentIR_cast* )
+  ---- ;; constructed object may not contain an abstract method
```
jaehyun1ee: Remove the comment.

spec/5-typing/5.11-typing-declaration.watsup
```diff
@@ -995,18 +997,18 @@ rulegroup Decl_ok/typedefDeclaration-typedef {
     ---- ;; if the type is monomorphic
     -- if $is_monomorphic_typeDefIR(typeDefIR)
     -- if typeIR = $typeIR_of_typeDefIR(typeDefIR)
-    ---- ;; create typedef and add to context
+    ---- ;; create typedef and add to context, keeping the created type bound
```
jaehyun1ee: Remove the ", keeping the created type bound".

spec/5-typing/5.11-typing-declaration.watsup
```diff
@@ -1018,12 +1020,12 @@ rulegroup Decl_ok/typedefDeclaration-typedef {
     ---- ;; if the type is polymorphic, yet has no type parameters
     -- if ~$is_monomorphic_typeDefIR(typeDefIR)
     -- if (eps, eps) = $typeParameterListIR_of_typeDefIR(typeDefIR)
-    ---- ;; create typedef and add to context
+    ---- ;; create typedef and add to context, keeping the created type bound
```
jaehyun1ee: Also applies here.

spec/5-typing/5.15.4-typing-callable-call.watsup
```diff
@@ -90,12 +90,18 @@ rule CallableType_ok/builtinMethodTypeIR-size:
...
   -- if typeIR_base = $type_of_typedExpressionIR(typedExpressionIR_base)
   -- if typeIR_base_unroll = $unroll_typeIR(typeIR_base)
   -- if ~(typeIR_base_unroll <: objectTypeIR)
   -- if ~(typeIR_base_unroll <: synthesizedTypeIR)
+  -- if ~(typeIR_base_unroll <: stringTypeIR)
+  -- if ~(typeIR_base_unroll <: errorTypeIR)
+  -- if ~(typeIR_base_unroll <: matchKindTypeIR)
+  -- if ~(typeIR_base_unroll <: intTypeIR)
+  -- if ~(typeIR_base_unroll <: listTypeIR)
+  -- if ~(typeIR_base_unroll <: simpleEnumTypeIR)
```
jaehyun1ee: Let's make a helper function instead: `$is_serializable_typeIR` to `2.2.3-type-aux.watsup`.

spec/6-dynamic-runtime/6.2.3-callable-callee.watsup
```diff
@@ -34,6 +34,9 @@ syntax builtinMethodCallee =
   BUILTIN_METHOD storageReference `. nameIR
     `< typeArgumentListIR > `( parameterListIR )
 
+syntax builtinSizeMethodCallee =
+  BUILTIN_SIZE_METHOD value
+
```
jaehyun1ee: Instead of adding `builtinSizeMethodCallee`, just add a rule to rule `Call_eval/builtinMethodCallee-size-expression`.

spec/7-instantiation/7.07-inst-declaration.watsup
```diff
@@ -520,7 +520,7 @@ rulegroup Decl_inst/typedefDeclarationIR-typedef {
     -- if $is_monomorphic_typeDefIR(typeDefIR)
     -- if typeIR = $typeIR_of_typeDefIR(typeDefIR)
     -- if aliasTypeDefIR = TYPEDEF nameIR typeIR
-    -- if IC_1 = $add_typeDef_i(GLOBAL, IC_0, nameIR, aliasTypeDefIR)
+    -- if IC_1 = $add_typeDef_i(GLOBAL, IC_local, nameIR, aliasTypeDefIR)
```
jaehyun1ee: Rename `IC_local` to `IC_1` and `IC_1` to `IC_2`.

spec/7-instantiation/7.07-inst-declaration.watsup
```diff
@@ -540,7 +540,7 @@ rulegroup Decl_inst/typedefDeclarationIR-typedef {
     -- if (eps, eps) = $typeParameterListIR_of_typeDefIR(typeDefIR)
     -- if typeIR = $specialize_typeDefIR(typeDefIR, eps)
     -- if typedefTypeIR = TYPEDEF nameIR typeIR
-    -- if IC_1 = $add_typeDef_i(GLOBAL, IC_0, nameIR, typedefTypeIR)
+    -- if IC_1 = $add_typeDef_i(GLOBAL, IC_local, nameIR, typedefTypeIR)
```
jaehyun1ee: Also applies here.

spec/7-instantiation/7.10.2-inst-control-declaration.watsup
```diff
@@ -22,10 +22,36 @@ rule ControlLocalDecl_inst/instantiationIR:
...
+  rule ControlLocalDecl_inst/variableDeclarationIR-init-unbound:
+    IC STO |- annotationList typeIR nameIR (`= typedExpressionIR) `;
+            : IC STO (annotationList typeIR nameIR (`= typedExpressionIR) `;)
+    ---- ;; initializer is dynamic (not compile-time known); defer to the
+    ---- ;; dynamic phase as before, without binding it for this pass
+    -- if DYN = $ctk_of_typedExpressionIR(typedExpressionIR)
+
+  rule ControlLocalDecl_inst/variableDeclarationIR-noinit:
+    IC STO |- annotationList typeIR nameIR eps `;
+            : IC STO (annotationList typeIR nameIR eps `;)
```
jaehyun1ee: This is an inadequate fix for #245. Should enforce the following constraint at the typing spec:

> The size is an optional property of a table. When present, its value must always be a compile-time known value that is an integer. (14.2.1.5)

spec/8-dynamic/8.03.1-eval-expression.watsup
```diff
@@ -334,7 +334,8 @@ rulegroup Expr_eval/memberAccessExpressionIR-typedExpressionIR {
     -- if HEADER_STACK `[ value_elem* `( _ ) ] = headerStackValue_base
     -- if nameIR = "size"
     -- if n_size = |value_elem*|
-    -- if expressionResult = `` (D n_size)
+    ---- ;; hs.size produces a 32-bit unsigned integer
```
jaehyun1ee: Delete the comment.

spec/8-dynamic/8.06-eval-statement.watsup
```diff
@@ -703,9 +703,11 @@ rulegroup Stmt_eval/forStatementIR-in {
     -- if typeIR_subst = $subst_type_e(LOCAL, EC_0, typeIR)
     -- if value_init = $default(typeIR_subst)
     -- if EC_3 = $add_var_e(LOCAL, EC_2, `` nameIR, value_init)
+    ---- ;; cast each element to the iterator type, as typing proved
```
jaehyun1ee: Delete the comment.

spec/8-dynamic/8.06-eval-statement.watsup
```diff
@@ -703,9 +703,11 @@ rulegroup Stmt_eval/forStatementIR-in {
...
+    -- if (value_cast = $cast_op(typeIR_subst, value))*
```
jaehyun1ee: This patch disagrees with the suggested fix in #235.

spec/8-dynamic/8.10.1-eval-call-callee.watsup
```diff
@@ -192,6 +192,23 @@ rulegroup Callee_eval/builtinMethodCallee-header-setInvalid {
...
+               : EC ARCH (`` builtinSizeMethodCallee)
+    ---- ;; if the method is min/maxSizeInBits/Bytes builtin
+    -- if nameIR <- [ "minSizeInBits", "minSizeInBytes",
+                      "maxSizeInBits", "maxSizeInBytes" ]
+    ---- ;; the size is derived from the receiver's static type
+    -- if _ `# `( typeIR_base _ ) = typedExpressionIR_base
+    -- if builtinSizeMethodCallee
+        = BUILTIN_SIZE_METHOD $sizeof(typeIR_base, nameIR)
+
+}
```
jaehyun1ee: Related to the comment on `builtinSizeMethodCallee`.

spec/8-dynamic/8.10.4-eval-call.watsup
```diff
@@ -586,13 +586,95 @@ rule Call_eval/builtinMethodCallee-header-union-isValid:
...
 
+;;;; syntax builtinSizeMethodCallee =
+;;;;   BUILTIN_SIZE_METHOD value
+
+rule Call_eval/builtinSizeMethodCallee:
+  p EC ARCH |- builtinSizeMethodCallee `@ `< eps > `( eps )
+             : EC ARCH returnResult
+  ---- ;; unpack the builtin size method callee
+  -- if BUILTIN_SIZE_METHOD value = builtinSizeMethodCallee
+  -- if returnResult = RETURN value
```
jaehyun1ee: Also, don't want this.

spec/7-instantiation/7.11.2-inst-constructor-call.watsup
```diff
@@ -118,17 +135,18 @@ rule Constructor_call/externConstructorDef:
...
   -- ExternMethods_inst:
       IC_callee_3 |- externMethodPrototypeIR* : IC_callee_4
   ---- ;; create an extern object with the final callee context
   -- if `{ (callableId `: externMethodDef)* } = IC_callee_4.BLOCK.CALLABLE
-  ---- ;; initialize extern state
+  ---- ;; initialize extern state, in declaration order; an omitted
+  ---- ;; @optional parameter has no callee-frame binding and is skipped
   -- if (_ _ _ id_cparam _ = constructorParameterIR)*
-  -- if (value_cparam = $find_map<id, value>(IC_callee_4.BLOCK.FRAME, id_cparam))*
+  -- if value_cparam* = $find_boundValues(IC_callee_4.BLOCK.FRAME, id_cparam*)
```
jaehyun1ee: This is strange. The `@optional` annotated-parameters should never reach here, since we perform:

```
    -- if GIVEN parameterIR_aligned*
          DEFAULT parameterIR_default*
        = $align_parameterListIR(
            parameterIR*, argumentIR*, id_default*, id_optional*)
    ---- ;; copy-in
    -- if p_shared = GLOBAL
    -- Copy_in:
        p_shared ARCH_0 |-
          p EC_0 parameterIR_aligned*
          `@ LOCAL EC_callee_3 argumentIR*
        ~> ARCH_1 EC_1 EC_callee_4 `# (`` storageReference?*)
```

where `parameterIR_aligned*` are the actually *given* ones.

p4spec/lib/interface/p4/parser.mly
jaehyun1ee: Done in 635a436d — comment removed, the fix kept.

p4spec/test/micro/programs/typedef_derived.p4
jaehyun1ee: Done — removed. That leaves the #251 parser fix without a test here; if you'd like, I can add a small positive witness under `testdata/regression` instead.

p4spec/test/micro/dune
jaehyun1ee: Done — removed both the @micro and @micro-det regression-wiring comments.

spec/5-typing/5.10-typing-statement.watsup
jaehyun1ee: Done — renamed to `TC_6` and shifted the following `TC_n` (`TC_6`→`TC_7`); no primed names left, in both `for` rules.

spec/5-typing/5.10-typing-statement.watsup
jaehyun1ee: Done — comment removed.

spec/5-typing/5.10-typing-statement.watsup
jaehyun1ee: Done — same rename + comment removal applied to the for-in rule.

spec/5-typing/5.11-typing-declaration.watsup
jaehyun1ee: Done — removed (the pre-existing identical comment on the object-initializer rule at ~:341 is left untouched).

spec/5-typing/5.11-typing-declaration.watsup
jaehyun1ee: Done — reverted to `;; create typedef and add to context`.

spec/5-typing/5.11-typing-declaration.watsup
jaehyun1ee: Done — both occurrences.

spec/7-instantiation/7.07-inst-declaration.watsup
jaehyun1ee: Done — `IC_local`→`IC_1`, `IC_1`→`IC_2`.

spec/7-instantiation/7.07-inst-declaration.watsup
jaehyun1ee: Done — same in the poly rule.

spec/8-dynamic/8.03.1-eval-expression.watsup
jaehyun1ee: Done — comment removed.

spec/8-dynamic/8.06-eval-statement.watsup
jaehyun1ee: Done — comment removed.

spec/5-typing/5.15.4-typing-callable-call.watsup
jaehyun1ee: Done — added `$is_serializable_typeIR` to `2.2.3-type-aux.watsup` (true exactly on the `$sizeof` domain: bool, int<w>, bit<w>, varbit, typedef/newtype, serializable enum, tuple, array, header stack, struct, header, header_union) and replaced the eight `<:` exclusions with a single `-- if $is_serializable_typeIR(typeIR_base)`. — 635a436d

spec/7-instantiation/7.10.2-inst-control-declaration.watsup
jaehyun1ee: Agreed — the control-local binding was the wrong direction. Reverted it and instead enforce a compile-time-known `size` at typing (14.2.1.5): in `TableProperty_ok/size`, `-- if DYN =/= $ctk_of_typedExpressionIR(typedExpressionIR)`. `size = n` (n a mutable local) is now rejected; a literal or `const` size still passes. The witness moved to the negative corpus, and it's rejected under `-det`. — 635a436d

spec/6-dynamic-runtime/6.2.3-callable-callee.watsup
jaehyun1ee: The catch: a size builtin's result depends only on the receiver's *static type*, and it's valid on an r-value receiver (`(expr).minSizeInBits()`). `builtinMethodCallee` carries a `storageReference` and its `Callee_eval` goes through `Lvalue_eval`, so folding size into it would get stuck on a size call over an r-value — a program that type-checks, i.e. a soundness gap. `builtinSizeMethodCallee` carries the computed size (from the static type) and needs no l-value, which is why it's separate. Happy to rename it or move the rule, but it can't reuse the `storageReference`-based `builtinMethodCallee` as-is. If size on r-value receivers is out of scope for you, I'll fold it in — let me know.

spec/8-dynamic/8.10.1-eval-call-callee.watsup
jaehyun1ee: Tied to the `builtinSizeMethodCallee` thread on `6.2.3` — this `Callee_eval` computes the size from the receiver's static type without needing an l-value, which is the point. Kept pending your call on whether r-value size receivers are in scope.

spec/8-dynamic/8.10.4-eval-call.watsup
jaehyun1ee: Same thread — this is the `Call_eval` rule that consumes `builtinSizeMethodCallee`. It stays as long as the dedicated callee does; see the reasoning on the `6.2.3` comment.

spec/8-dynamic/8.06-eval-statement.watsup
jaehyun1ee: The suggested fix in #235 casts the collection to a `LIST` of the iterator type at typing via `$cast_unary(..., LIST<typeIR_var>)`, but there is no `LIST<a> -> LIST<b>` clause in `Cast_impl` (a for-in collection already has a `LIST` type, not `SEQ`), so that diff doesn't typecheck against the current tree. This patch performs the same proven per-element conversion in the evaluator via `$cast_op`, reusing the elementwise-cast idiom already in `$cast_sequence`. If you'd rather do it typing-side, I can add the missing `LIST -> LIST` cast rule instead — your call.

spec/7-instantiation/7.11.2-inst-constructor-call.watsup
jaehyun1ee: `$align_parameterListIR_given` emits the given params in **argument** order, not declaration order: for a named/reordered call it resolves each param by name but walks the position list, so the aligned list follows `argumentIR*` (`4.1-ir-call-helper.watsup:106-117`). The native positional extern init (`object.ml`, e.g. `Counter.init`) reads values in **declaration** order, so this builds the value list from `constructorParameterIR*` (declaration order) with a frame lookup — an omitted `@optional` simply has no frame binding and is skipped. Using `parameterIR_aligned*` directly reintroduces the argument-order crash on `counter(type=…, size=…)` (the `issue-232b` regression). Open to a cleaner formulation if you have one.

spec/5-typing/5.14.1-typing-control-table.watsup
```diff
@@ -885,6 +885,8 @@ rulegroup TableProperty_ok/size-and-others {
     -- if typeIR_unroll = $unroll_typeIR(typeIR)
     -- if (typeIR_unroll <: integerTypeIR)
           /\ ~(typeIR_unroll <: varBitTypeIR)
+    ---- ;; the size must be a compile-time known value
```
jaehyun1ee: Delete the comment.

spec/5-typing/5.04-typing-relation.watsup
jaehyun1ee, line 1: Adding `typingContext` to the `Parameter_ok` signature is *a* fix, but I don't like it.

Consider changing it as:

```
relation Parameter_ok:
  cursor typingContext |- parameter : parameterIR `# typeId*
```

so `Parameter_ok` only *checks*, and doesn't update the context, where `ParameterList_ok` is responsible for the typing context update.

spec/8-dynamic/8.06-eval-statement.watsup
jaehyun1ee: In that case, let's defer the fix. Doesn't seem like a trivial fix to me.

spec/6-dynamic-runtime/6.2.3-callable-callee.watsup
jaehyun1ee: Ouch, then let's defer the fix. If we are to add `builtinSizeMethodCallee`, I would like a consistent treatment of other builtins, having `builtinSOMEMethodCallee` per each builtin in P4. Consider opening an issue about it.

spec/5-typing/5.14.1-typing-control-table.watsup
jaehyun1ee: Done — comment removed. — 970f1e41

spec/5-typing/5.04-typing-relation.watsup
jaehyun1ee: Done — `Parameter_ok` is now a pure check: `cursor typingContext |- parameter : parameterIR `#` typeId*`, no context update. `Parameters_ok` performs the `$add_parameter_t` fold (and still passes the pre-list context to each `Parameter_ok`, so a default can't alias a sibling; duplicate-name detection stays in the fold). — 970f1e41

spec/8-dynamic/8.06-eval-statement.watsup
jaehyun1ee: Deferred — reverted the for-in element-cast patch and dropped its witness; #235 stays open. — 970f1e41

spec/6-dynamic-runtime/6.2.3-callable-callee.watsup
jaehyun1ee: Deferred — reverted the size-builtin fix (receiver restriction, runtime size-method callee, and the `$is_serializable_typeIR` helper) and dropped the two witnesses. Opened #255 to track a consistent per-builtin `builtin<Name>MethodCallee` scheme; it covers #225 and #226. — 970f1e41

spec/7-instantiation/7.11.2-inst-constructor-call.watsup
jaehyun1ee: Done, implemented your approach in 72e3ea8b — `$init_objectState` is now name-based. It gains an `id*` argument; the rule passes `id_cparam*` + `value_cparam*` for **given ++ default** params (`Copy_in_inst_default` binds defaults in the frame, so aligned-only would drop them; an omitted `@optional` has no frame binding). A `named_args` helper (`spec_impl/unpack.ml`) pairs ids with values, and every extern `init` (v1model ×4, ebpf ×1, psa ×5) plus each `eval_extern_init` dispatcher looks its constructor arguments up by name. `$find_boundValues` deleted; ordering is irrelevant end to end. Verified: #232 + #232b pass incl `-det`; full SL sim suite green across v1model/ebpf/psa.

spec/6-dynamic-runtime/6.4.1-object.watsup
```diff
@@ -12,7 +12,7 @@
 extern syntax objectState
 syntax externMethodDefEnv = map<callableId, externMethodDef>
 
-extern dec $init_objectState(nameIR, typeIR*, value*) : objectState
+extern dec $init_objectState(nameIR, typeIR*, id*, value*) : objectState
   hint(prose_in "initialization of" %0#`<%1`>#`(%2`) "in the target")
```
jaehyun1ee: Update the hint.

spec/7-instantiation/7.11.2-inst-constructor-call.watsup
```diff
@@ -118,18 +114,22 @@ rule Constructor_call/externConstructorDef:
...
   -- ExternMethods_inst:
       IC_callee_3 |- externMethodPrototypeIR* : IC_callee_4
   ---- ;; create an extern object with the final callee context
   -- if `{ (callableId ':' externMethodDef)* `} = IC_callee_4.BLOCK.CALLABLE
-  ---- ;; initialize extern state
-  -- if (_ _ _ id_cparam _ = constructorParameterIR)*
-  -- if (value_cparam = $find_map<id, value>(IC_callee_4.BLOCK.FRAME, id_cparam))*
-  -- if objectState = $init_objectState(nameIR, typeArgumentIR*, value_cparam*)
+  ---- ;; initialize extern state from the bound constructor arguments,
+  ---- ;; keyed by parameter name; an omitted @optional has no binding
```
jaehyun1ee: Just put `;; initialize extern state`.

p4spec/lib/backend-sim/spec_impl/unpack.ml
```diff
@@ -5,6 +5,16 @@ module Value = Runtime.Value
 
 let first fs x = List.find_map (fun f -> f x) fs
 
+(* Pair a list of parameter-name ids with a list of argument values, so an
+   extern init can look its constructor arguments up by name rather than by
+   position. *)
```
jaehyun1ee: Just put `(* Pair a list of parameter-name ids *)`.

p4spec/lib/backend-sim/spec_impl/unpack.ml
```diff
@@ -5,6 +5,16 @@ module Value = Runtime.Value
...
+
+let named_args (value_ids : Value.t) (value_args : Value.t) :
```
jaehyun1ee: Rename to `assoc_args`.

p4spec/lib/backend-sim/spec_impl/unpack.ml
```diff
@@ -5,6 +5,16 @@ module Value = Runtime.Value
...
+    (string * Value.t) list =
+  List.combine
+    (List.map Value.Get.text (Value.Get.list value_ids))
+    (Value.Get.list value_args)
```
jaehyun1ee: Reorder: This should come before `first`.

spec/5-typing/5.09-typing-parameter.watsup
jaehyun1ee, line 1: Don't update the typing context in either `Parameter_ok` or `Parameters_ok`. Instead, update at `ParameterList_ok`.

```
rule ParameterList_ok:
  p TC_0 |- parameterList
          : TC_1 parameterIR* '#' typeId_fresh*
  -- if parameter* = $flatten_parameterList(parameterList)
  -- Parameters_ok:
      p TC_0 |- parameter*
              : parameterIR* '#' typeId_fresh*
  -- if (_ _ _ id_param _ = parameterIR)*
  -- if TC_1 = $add_parameters_t(p, TC_0, parameterIR*)
```

jaehyun1ee: Done — neither `Parameter_ok` nor `Parameters_ok` touches the context now; both are pure checks. `ParameterList_ok` does the update via a new `$add_parameters_t(p, TC_0, parameterIR*)` fold (defined next to `$add_parameter_t`), and `ConstructorParameterListOpt_ok` does the same. Duplicate-name detection still falls out of the fold (`$add_var_t`'s `~$in_set`), and since every parameter is checked under the same pre-update context a default still can't alias a sibling. — d422aadd

spec/6-dynamic-runtime/6.4.1-object.watsup
jaehyun1ee: Done — the added `id*` argument shifted `value*` to `%3`; the hint now reads `%0#`<%1`>#`(%2 '=' %3`)`, rendering the arguments as `name = value` pairs. — d422aadd

spec/7-instantiation/7.11.2-inst-constructor-call.watsup
jaehyun1ee: Done — now just `;; initialize extern state`. — d422aadd

p4spec/lib/backend-sim/spec_impl/unpack.ml
jaehyun1ee: Done — `(* Pair a list of parameter-name ids *)`. — d422aadd

p4spec/lib/backend-sim/spec_impl/unpack.ml
jaehyun1ee: Done — renamed to `assoc_args` (definition + all 10 call sites). — d422aadd

p4spec/lib/backend-sim/spec_impl/unpack.ml
jaehyun1ee: Done — moved above `first`. — d422aadd

PR #254, opened 2026-07-28
Add table groups

jaehyun1ee:
## Summary

Revives table groups (#117): a `tblgroup` batch-declares N single-argument table functions ("columns") that share one signature and namespace, and renders them as one matrix — rows are the input patterns, columns are the functions, cells are the results.

```
tblgroup nestable_synthesized(typeIR) : bool
  hint(prose_true  %0 "can be nested in a"    %1)
  hint(prose_false %0 "cannot be nested in a" %1)
{
  tbldec $set              hint(tbl_col "set type")
  tbldec $tuple_in_set     hint(tbl_col "tuple nested in set type")
  tbldec $sequence_in_set  hint(tbl_col "sequence nested in set type")
}

tbldef $nestable_synthesized.set =
  | boolTypeIR         => true
  | TUPLE `< typeIR* > => $forall_($nestable_synthesized.tuple_in_set(typeIR)*)
  | ...
```

This is landed in phases so each is separately reviewable and build/test-green. The plan below is the whole arc; this PR currently contains the foundation (phases 1 through 6). Later phases are pushed to the same branch as they land.

## Plan

**Phase 1 — `refine_rows`.** The row-refinement primitive the matrix uses: the meet of the columns' pattern partitions (the coarsest set of non-overlapping rows that preserves every distinction any column makes). Lands ahead of its use. *(in this PR, unwired)*

**Phase 2 — demote binary compat tables to functions.** The ten `(typeIR, typeIR)` compatibility relations are not single-parameter tables; they become `dec`/`def` functions, so "every table is single-argument" holds. *(in this PR)*

**Phase 3 — table keywords `tbldec`/`tbldef`.** Two-token `tbl dec`/`tbl def` become single tokens, peer to the function keywords `dec`/`def`. *(in this PR)*

**Phase 4 — higher-order table parameter `tbldef`.** A function can receive a table as an argument, mirroring the function `def` parameter. *(in this PR)*

**Phase 5 — `Table` becomes a peer sort of `Func`.** `Table` leaves `Func.t`; runtime gains a `Table` module + `TEnv` and a `TableV` value. *(in this PR)*

**Phase 6 — split `Table` from `Func` in the static tower and elaborator.** Removes the last `Func.Table`. *(in this PR)*

**Phase 7 — table-group syntax and semantics.** `tblgroup` block + `$g.col` addressing in the parser and AST; elaboration collects the columns and dup-checks names; structuring builds the executable SL matrix — `refine_rows` across the columns, then cross-column antiunification so a shared row header and its cells agree on their bound variables; runtime executes `$g.col(x)` against the matrix. Singleton tables become one-column groups. *(upcoming)*

**Phase 8 — render the matrix.** SL matrix to prose: rows, columns, and cells to an AsciiDoc table. *(upcoming)*

**Phase 9 — render polish.** Legend for complex cell expressions, largest row as "otherwise", corner arrows from `prose_true`/`prose_false`, column headers from `tbl_col`. *(upcoming)*

**Phase 10 — migrate real tables to groups.** Convert `is_defaultable`/`equalable`/`assignable`, `compat_table_key`, and the `nestable_*` family to `tblgroup`s, one per commit, verifying exclusiveness against the spec. *(upcoming)*

## What is in this PR

**1. Demote binary compat tables to functions.**

```
;; before                                     ;; after
tbldec $compat_shift(typeIR, typeIR) : bool   dec $compat_shift(typeIR, typeIR) : bool
tbldef $compat_shift =                        def $compat_shift(INT, INT) = true
  | (INT, INT) => true                        def $compat_shift(INT `< _ >, BIT `< _ >) = true
  | (INT `< _ >, BIT `< _ >) => true          def $compat_shift(_, _) = false
  | (_, _) => false
```

**2. Table keywords `tbldec`/`tbldef`,** peer to `dec`/`def`.

**3. Higher-order table parameter `tbldef`.** `$reduce_serenum_unary` — the only helper that receives a table value — takes a table parameter, and its call sites pass `compat_lnot`, `compat_bnot`, `compat_uplusminus`, `compat_array_index`, `compat_bitslice_*` as `tbldef` arguments.

```
;; function param / arg           ;; table param / arg
def $check(typeIR) : bool          tbldef $check(typeIR) : bool
... def $compat_lnot ...           ... tbldef $compat_lnot ...
```

**4. `Table` is a peer sort of `Func`.** Each runtime tower gains a `Table` module and a `TEnv` peer to `FEnv`, a passed table is a `TableV` value peer to `FuncV`, and the elaborator stores tables in a `tenv` separate from `fenv`.

```ocaml
(* before: dynamic-*/func.ml *)          (* after *)
type t =                                 type t = Extern ... | Builtin ... | Defined ...
  | Extern ... | Builtin ...             (* new table.ml *)
  | Table of param list * typ * ...      module Table = struct
  | Defined ...                            type t = param list * typ * tablerow list end
                                         (* new envs.ml *)
                                         module TEnv = MakeFIdEnv (Table)
```

Call dispatch is transitional: a bare call resolves in the function environment first, then the table environment. A distinct table-call node is left for Phase 7.

**5. `refine_rows`** in the `Pattern` module (unwired; used in Phase 7).

## Bugs fixed along the way

Removing `Table` from `Func` broke three call-resolution sites that assumed a table was a function. Each surfaced as `function X is undefined`, or, worse, a non-terminating fixpoint.

**`Il.Eq.eq_arg`** had no `TableA` arm, so two blocks holding the same table argument compared unequal, and the structuring optimizer's fixpoint (`optimize_loop` stops when `eq_block old new` holds) never converged — an infinite loop, not an error.

```diff
   | DefA id_a, DefA id_b -> eq_id id_a id_b
+  | TableA id_a, TableA id_b -> eq_id id_a id_b
   | _ -> false
```

**Table tail-call** (`interp-sl`): `eval_return_instr` used a strict `find_func` to decide tail-call optimization, so `Return $nestable_tuple(typeIR)` — a table in tail position — raised `function nestable_tuple is undefined`. A table callee now invokes directly instead of taking the function tail-call path.

**Elaborator call resolution**: `find_func_signature`, used to type-check every `CallE`, looked only in `fenv`; a table call inside a table body (`$is_defaultable_typeIR(typeIR)`) resolved to nothing. It now falls back to the table env for a table's monomorphic signature.

## Verification

`make build` and `make test-fast` (`@speclang @p4parse @run-sl @sim-sl`) green at HEAD and at each of the six commits. Golden `elab/algo/struct/annotate.expected` change only where a table's rendering or the `tbldef` parameter appears; the `Table`-sort split (phases 5 and 6) is elaboration-output-preserving and leaves the goldens unchanged. `make test-boot` and the `-al`/`-pl` runtime towers beyond `@run-sl`/`@sim-sl` were not run.


p4spec/lib/interface/spectec/ali/boot.ml
```diff
@@ -21,6 +21,16 @@ let rec boot_param (param : Il.param) : Value.t =
         mop_def_param
         <|! [ value_id; value_tparams; value_params; value_typ ]
         <<|! typ_param <<<| at)
+  (* A table is still a `Func.t` in this commit: boot it as a def parameter. *)
```
jaehyun1ee: Remove the comment.

p4spec/lib/interface/spectec/common/boot.ml
```diff
@@ -351,6 +353,10 @@ and boot_arg (arg : Il.arg) : Value.t =
   | DefA id ->
       let value_id = boot_id id in
       Value.Make.(mop_def_arg <|! [ value_id ] <<|! typ_arg <<<| at)
+  (* A table is still a `Func.t` in this commit: boot it as a def argument. *)
```
jaehyun1ee: Remove the comment.

p4spec/lib/interface/spectec/sli/boot.ml
```diff
@@ -23,6 +23,16 @@ let rec boot_param (param : Sl.param) : Value.t =
         mop_def_param
         <|! [ value_id; value_tparams; value_params; value_typ ]
         <<|! typ_param <<<| at)
+  (* A table is still a `Func.t` in this commit: boot it as a def parameter. *)
```
jaehyun1ee: Remove the comment.

p4spec/lib/interp/interp-al/ctx.ml
```diff
@@ -197,10 +211,28 @@ module Make () = struct
...
       (tparam list * typ list * typ) option =
-    find_func_opt ctx fid
-    |> Option.map (fun (_, func) -> Func.get_signature func)
+    match find_func_opt ctx fid with
+    | Some (_, func) -> Some (Func.get_signature func)
+    | None ->
+        find_table_opt ctx fid
+        |> Option.map (fun (_, table) ->
+               let typs, typ = Table.get_signature table in
+               ([], typs, typ))
```
jaehyun1ee: Should have dedicated functions for each: `find_func_signature_opt` and `find_table_signature_opt`.

p4spec/lib/interp/interp-pl/ctx.ml
```diff
@@ -236,10 +245,33 @@ module Make () = struct
...
       (tparam list * typ list * typ) option =
-    find_func_opt ctx fid
-    |> Option.map (fun (_, func) -> Func.get_signature func)
+    match find_func_opt ctx fid with
+    | Some (_, func) -> Some (Func.get_signature func)
+    | None ->
+        find_table_opt ctx fid
+        |> Option.map (fun (_, table) ->
+               let typs, typ = Table.get_signature table in
+               ([], typs, typ))
```
jaehyun1ee: Also applies here.

p4spec/lib/interp/interp-sl/ctx.ml
```diff
@@ -236,10 +245,33 @@ module Make () = struct
...
       (tparam list * typ list * typ) option =
-    find_func_opt ctx fid
-    |> Option.map (fun (_, func) -> Func.get_signature func)
+    match find_func_opt ctx fid with
+    | Some (_, func) -> Some (Func.get_signature func)
+    | None ->
+        find_table_opt ctx fid
+        |> Option.map (fun (_, table) ->
+               let typs, typ = Table.get_signature table in
+               ([], typs, typ))
```
jaehyun1ee: Also applies here.

p4spec/lib/lang/il/print.ml
```diff
@@ -428,7 +435,7 @@ let rec string_of_def def =
       "builtin def " ^ string_of_defid defid ^ string_of_tparams tparams
       ^ string_of_params params ^ " : " ^ string_of_typ typ
   | TableDecD (defid, params, typ, tablerows, _) ->
-      "tbl def " ^ string_of_defid defid ^ string_of_params params ^ " : "
+      "tbldef " ^ string_of_defid defid ^ string_of_params params ^ " : "
       ^ string_of_typ typ ^ " ="
       ^ string_of_tablerows tablerows
```
jaehyun1ee: Overall, reorder `TableDecD` cases in type definition and match-cases. The order should be: `ExternDecD`, `BuiltinDecD`, `FuncDecD`, and `TableDecD`. Also applies for EL/AL/SL/PL.

p4spec/lib/pass/elaborate/ctx.ml
```diff
@@ -189,13 +191,19 @@ let bound_defined_func (ctx : t) (fid : FId.t) : bool =
...
+         | Func.Extern (tparams_il, params_il, typ_il)
+         | Func.Builtin (tparams_il, params_il, typ_il)
+         | Func.Defined (tparams_il, params_il, typ_il, _, _) ->
+             (tparams_il, params_il, typ_il))
+  with
+  | Some signature -> Some signature
+  (* a table call resolves to its signature; tables are monomorphic *)
+  | None ->
+      find_table_func_opt ctx fid
+      |> Option.map (fun (params_il, typ_il, _) -> ([], params_il, typ_il))
```
jaehyun1ee: Here as well.

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1188,6 +1188,11 @@ and elab_param (ctx : Ctx.t) (param : param) : Il.param =
       let params_il = List.map (elab_param ctx_local) params in
       let typ_il = elab_plaintyp ctx_local plaintyp in
       Il.DefP (id, tparams, params_il, typ_il) $ param.at
+  | TableP (id, params, plaintyp) ->
+      (* Tables are monomorphic: no type parameters. *)
```
jaehyun1ee: Delete the comment.

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1229,12 +1234,43 @@ and elab_arg ?(as_def = false) (ctx : Ctx.t) (param_il : Il.param) (arg : arg) :
            (Id.to_string id_p));
       let arg_il = Il.DefA id_a $ arg.at in
       (ctx, arg_il)
+  | TableP (id_p, params_il_p, typ_il_p), TableA id_a when as_def ->
+      check (id_p.it = id_a.it) arg.at
+        (F.asprintf
+           "table argument does not match the declared table parameter %s"
+           (Id.to_string id_p));
+      (* A table is registered in the peer table env, not among functions. *)
```
jaehyun1ee: Delete the comment.

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1229,12 +1234,43 @@ and elab_arg ?(as_def = false) (ctx : Ctx.t) (param_il : Il.param) (arg : arg) :
...
+      let ctx = Ctx.add_table_func_dec ctx id_p params_il_p typ_il_p in
```
jaehyun1ee: Overall, I don't want to name `table_func`. These are now just `table`s.

p4spec/lib/runtime/value/match.ml
```diff
@@ -79,6 +79,14 @@ let rec sub_ (find_typdef_opt : TId.t -> Type.Typdef.t option)
           Type.Equiv.equiv_functyp find_typdef_opt typ.at tparams_t
             typs_params_t typ_ret_t tparams_v typs_params_v typ_ret_v
       | _ -> false)
+  | TableT (typs_params_t, typ_ret_t) -> (
+      (* Tables are monomorphic: compare signatures with empty tparams. *)
```
jaehyun1ee: Remove the comment.

p4spec/lib/runtime/value/match.ml
```diff
@@ -79,6 +79,14 @@ let rec sub_ (find_typdef_opt : TId.t -> Type.Typdef.t option)
...
+      match value.it with
+      | TableV fid ->
+          let tparams_v, typs_params_v, typ_ret_v = find_func fid in
+          Type.Equiv.equiv_functyp find_typdef_opt typ.at [] typs_params_t
```
jaehyun1ee: Maybe rename `equiv_functyp` to `equiv_arrowtyp`?

p4spec/lib/interp/interp-al/interp.ml
```diff
@@ -1387,19 +1404,30 @@ module Make (Interface : Run.INTERFACE) (Extern : Run.EXTERN) () :
...
+        | _ -> false)
       values_input
 
   and invoke_func ?(internal : bool = true) (ctx : Ctx.t) (id : id)
       (targs : targ list) (values_input : value list) : value backtrack =
     Hook.on_func_enter id values_input;
-    (* Find the function *)
-    let cursor, func = Ctx.find_func ctx id in
+    (* Find the callee, either a function or (transitionally) a table.
+       A later elaborator change emits a distinct table-call node. *)
```
jaehyun1ee: Just put `(* Find the callee *)`.

p4spec/lib/interface/spectec/ali/boot.ml
jaehyun1ee: Removed (9364a025).

p4spec/lib/interface/spectec/common/boot.ml
jaehyun1ee: Removed (9364a025).

p4spec/lib/interface/spectec/sli/boot.ml
jaehyun1ee: Removed (9364a025).

p4spec/lib/interp/interp-al/ctx.ml
jaehyun1ee: Split: `find_func_signature_opt` is now function-only, and a dedicated `find_table_signature_opt` resolves tables. The bare-call resolver (`find_func_signature`) tries the function env then the table env. (9364a025)

p4spec/lib/interp/interp-pl/ctx.ml
jaehyun1ee: Done — same split in interp-pl. (9364a025)

p4spec/lib/interp/interp-sl/ctx.ml
jaehyun1ee: Done — same split in interp-sl. (9364a025)

p4spec/lib/lang/il/print.ml
jaehyun1ee: Reordered: `TableDecD` now comes last (ExternDecD, BuiltinDecD, FuncDecD, TableDecD) in the `def'` definition and every match, across EL/IL/AL/SL/PL. (9364a025)

p4spec/lib/pass/elaborate/ctx.ml
jaehyun1ee: Done — `find_func_signature_opt` function-only + `find_table_signature_opt` here too, and `infer_call_exp` tries both. (9364a025)

p4spec/lib/pass/elaborate/elab.ml
jaehyun1ee: Deleted. (9364a025)

p4spec/lib/pass/elaborate/elab.ml
jaehyun1ee: Deleted. (9364a025)

p4spec/lib/pass/elaborate/elab.ml
jaehyun1ee: Renamed: `add_table_func_dec`->`add_table_dec`, `add_table_func_tablerows`->`add_table_rows`, `find_table_func`->`find_table`, `find_table_func_opt`->`find_table_opt`. (9364a025)

p4spec/lib/runtime/value/match.ml
jaehyun1ee: Removed. (9364a025)

p4spec/lib/runtime/value/match.ml
jaehyun1ee: Renamed to `equiv_arrowtyp` (it compares both function and table signatures). (9364a025)

p4spec/lib/interp/interp-al/interp.ml
jaehyun1ee: Done — now just `(* Find the callee *)`. (9364a025)

p4spec/lib/interp/interp-al/ctx.ml
jaehyun1ee: Made the raising pair dedicated too: `find_func_signature` is function-only, `find_table_signature` is table-only (no merged fallback in either). `Value.Match.sub`/`subs` now take both a `find_func` and a `find_table` callback and use each for `FuncV`/`TableV`; `eval_arg`'s `TableA` arm resolves via `find_table_signature`. (944fb567)

p4spec/lib/interp/interp-pl/ctx.ml
jaehyun1ee: Raising pair made dedicated too here: `find_func_signature` function-only, `find_table_signature` table-only (no fallback). The try-both is at the call site, and `Value.Match.sub`/`subs` take both callbacks. (944fb567)

p4spec/lib/interp/interp-sl/ctx.ml
jaehyun1ee: Raising pair made dedicated too here: `find_func_signature` function-only, `find_table_signature` table-only (no fallback). The try-both is at the call site, and `Value.Match.sub`/`subs` take both callbacks. (944fb567)

p4spec/lib/pass/elaborate/ctx.ml
jaehyun1ee: Elaborate too: `find_func_signature` function-only, added dedicated `find_table_signature`; `infer_call_exp` tries `find_func_signature_opt` then `find_table_signature_opt`. (944fb567)

p4spec/lib/interp/interp-al/interp.ml
```diff
@@ -1387,19 +1409,74 @@ module Make (Interface : Run.INTERFACE) (Extern : Run.EXTERN) () :
...
+                invoke_body ())
+            in
+            Hook.on_func_exit id;
+            result
+            |> back_nest id.at (fun () ->
+                   F.asprintf "invocation of function %s%s failed"
+                     (Il.Print.string_of_defid id)
+                     (Il.Print.string_of_targs targs))
+        | None ->
+            error id.at (F.asprintf "function or table `%s` is undefined" id.it))
```
jaehyun1ee: `invoke_arrow` should be left as a simple dispatcher. This logic should exist in `invoke_table`. Applies elsewhere in AL/PL/SL interpreters.

jaehyun1ee (commented):
Also, I'd like the `eval_func` API to be renamed to `eval_arrow`.

p4spec/lib/interp/interp-al/interp.ml
jaehyun1ee: Done — `invoke_arrow` is now find + divert only. The table find/caching/hooks moved into `invoke_table`, and the raw execution is `invoke_table_body` (mirroring `invoke_func`/`invoke_func_body`). Applied to SL/AL/PL. (ec891a34)

p4spec/lib/lang/el/ast.ml
```diff
@@ -174,6 +177,10 @@ and rule' = id * id * exp * prem list
 and tablerow = tablerow' phrase
 and tablerow' = exp * exp
 
+(* column declaration: id hint* *)
```
jaehyun1ee: Delete the comment.

p4spec/lib/lang/el/ast.ml
```diff
@@ -196,14 +203,18 @@ and def' =
...
   (* `dec` id `<` list(tparam, `,`) `>` list(param, `,`) `:` plaintyp hint* *)
   | FuncDecD of id * tparam list * param list * plaintyp * hint list
-  (* `tbl` `def` id `=` list(`|` tablerow, nl) *)
-  | TableDefD of id * tablerow list
+  (* `tbl` `dec` id list(param, `,`) `:` plaintyp hint* *)
+  | TableDecD of id * param list * plaintyp * hint list
+  (* `tblgroup` id `(` param `)` `:` plaintyp hint* `{` tablecol* `}` *)
+  | TableGroupD of id * param * plaintyp * tablecol list * hint list
   (* `def` id `<` list(tparam, `,`) `>` list(arg, `,`) `=` exp list(`--` prem, nl) *)
   | FuncDefD of id * tparam list * arg list * exp * prem list
```
jaehyun1ee, line 208: Reorder: this should come after `FuncDecD`.

p4spec/lib/lang/el/print.ml
```diff
@@ -286,19 +290,28 @@ let string_of_def def =
...
+      ^ string_of_plaintyp plaintyp
+  | TableGroupD (defid, param, plaintyp, cols, _hints) ->
+      "tblgroup " ^ string_of_defid defid ^ "(" ^ string_of_param param
+      ^ ") : " ^ string_of_plaintyp plaintyp ^ " {\n  "
+      ^ String.concat "\n  "
+          (List.map (fun col -> let id, _ = col.it in "tbldec " ^ id.it) cols)
+      ^ "\n}"
   | FuncDefD (defid, tparams, args, exp, prems) ->
       "def " ^ string_of_defid defid ^ string_of_tparams tparams
       ^ string_of_args args ^ " = " ^ string_of_exp exp ^ string_of_prems prems
```
jaehyun1ee: Also reorder.

p4spec/lib/lang/el/render.ml
```diff
@@ -298,8 +299,10 @@ let render_def def =
       render_rulegroup_def id_rel id_rulegroup rules
   | ExternDecD _ -> ""
   | BuiltinDecD _ -> ""
-  | TableDecD _ -> ""
   | FuncDecD _ -> ""
-  | TableDefD _ -> ""
+  | TableDecD _ -> ""
+  | TableGroupD _ -> ""
   | FuncDefD _ -> ""
```
jaehyun1ee, line 303: Reorder here as well.

p4spec/lib/pass/elaborate/ctx.ml
```diff
@@ -25,6 +25,17 @@ let fresh () =
...
+(* Table groups *)
+
+(* A group holds its single parameter, its return type, its columns (id +
+   hints), and each column's rows (keyed by column id, empty at decl time). *)
+type group = {
+  param : Il.param;
+  typ : Il.typ;
+  cols : (Id.t * Il.hint list) list;
+  rows : Il.tablerow list IdMap.t;
+}
```
jaehyun1ee: This better belongs in `Runtime.*.Table`.

p4spec/lib/pass/elaborate/ctx.ml
```diff
@@ -40,6 +51,10 @@ type t = {
   renv : REnv.t;
   (* Map from function ids to functions *)
   fenv : FEnv.t;
+  (* Map from table ids to tables *)
+  tenv : TEnv.t;
```
jaehyun1ee, line 44: Likewise `rulegroup`s, can we treat simple `tbl` as a singleton table group?

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1688,9 +1735,44 @@ and elab_tablerows (ctx : Ctx.t) (at : region) (id : id)
...
-  let params_il, typ_il, _ = Ctx.find_table_func ctx id in
+  let params_il, typ_il, _ = Ctx.find_table ctx id in
   let tablerows_il = elab_tablerows ctx at id params_il typ_il tablerows in
-  Ctx.add_table_func_tablerows ctx id tablerows_il
+  Ctx.add_table_rows ctx id tablerows_il
+
+(* Elaboration of table group declarations and definitions *)
+
+and elab_table_group_dec_def (ctx : Ctx.t) (at : region) (id : id)
+    (param : param) (plaintyp : plaintyp) (cols : tablecol list)
```
jaehyun1ee: Violates the coding convention. Put `tablecols : tablecol list`.

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1688,9 +1735,44 @@ and elab_tablerows (ctx : Ctx.t) (at : region) (id : id)
...
+  let typ_il = elab_plaintyp ctx plaintyp in
+  check (typ_il.it = Il.BoolT) typ_il.at "table group must return a boolean type";
+  let cols =
+    List.map (fun (col : tablecol) -> col.it) cols
+  in
+  check
+    (List.map fst cols |> List.map it |> distinct ( = ))
+    at "table group columns are not distinct";
+  let ctx = Ctx.add_group_dec ctx id param_il typ_il cols in
+  let cols_il =
```
jaehyun1ee: Should be named `tablecols_il`.

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1688,9 +1735,44 @@ and elab_tablerows (ctx : Ctx.t) (at : region) (id : id)
...
+    List.map (fun (cid, chints) -> (cid, [], chints) $ cid.at) cols
```
jaehyun1ee: `hints` instead of `chints`.

p4spec/lib/lang/el/ast.ml
jaehyun1ee: Deleted. (024e7c7b)

p4spec/lib/lang/el/ast.ml
jaehyun1ee: Reordered so each sort's dec/def is adjacent: FuncDecD, FuncDefD, TableDecD, TableDefD, TableGroupD, TableGroupDefD — in the EL `def'`, and the print/render matches. (024e7c7b)

p4spec/lib/lang/el/print.ml
jaehyun1ee: Done, same order. (024e7c7b)

p4spec/lib/lang/el/render.ml
jaehyun1ee: Done, same order. (024e7c7b)

p4spec/lib/pass/elaborate/ctx.ml
jaehyun1ee: Moved to `Runtime.Static.Table` as `Table.group`; the elaborate ctx now stores `Table.group IdMap.t`. (024e7c7b)

p4spec/lib/pass/elaborate/ctx.ml
jaehyun1ee: Yes — that's the intended end state ("everything is a group": a standalone table desugars to a one-column group, retiring the separate tenv/TableDecD path, just as a rule always lives in a rulegroup). Planned as its own phase after the group pipeline (structure/render/runtime) lands, so the desugar + legacy-path removal come as one reviewable golden-regen step. Happy to pull it forward if you prefer.

p4spec/lib/pass/elaborate/elab.ml
jaehyun1ee: Renamed the parameter to `tablecols : tablecol list`. (024e7c7b)

p4spec/lib/pass/elaborate/elab.ml
jaehyun1ee: Renamed to `tablecols_il`. (024e7c7b)

p4spec/lib/pass/elaborate/elab.ml
jaehyun1ee: Renamed to `hints`. (024e7c7b)

jaehyun1ee:
## Design discussion: how a table group matrix executes

Structuring builds one matrix per group (`refine_rows` → refined rows × columns, each cell `= out_exp * block`), and `$g.col(x)` runs a column by concatenating that column's cell blocks in row order (first match returns), reusing `invoke_table_body`. A throwaway harness (`eval_func` on nullary callers) found this executes incorrectly:

```
tblgroup $g(typ) : bool { tbldec $a  tbldec $b }
tbldef $g.a = | BOOL => true | _ => false
tbldef $g.b = | INT  => true | _ => false

$g.a(BOOL)=true  $g.a(INT)=false  $g.a(NAMED)=false   ; correct
$g.b(BOOL)=false $g.b(INT)=false  ...                  ; WRONG — $g.b(INT) must be true
```

### Root cause

`refine_rows` imposes one global row order (`BOOL; INT; NAMED`) across all columns. `totalize_column` expands each column's `| _` closer into per-complement cells, but those synthesized cells carry the closer's premises verbatim — which are empty — so they emit an **unguarded** return:

```
;; column $b, cell blocks in row order
[$b]@BOOL -> Return false            ;; from `_` closer, no guard — swallows every input
[$b]@INT  -> if typ matches INT then Return true   ;; never reached
```

For `$a` the unguarded `false` cells land after the guarded `BOOL` cell, so `$a` is right by ordering; `$b`'s explicit row is `INT`, which the earlier unguarded `BOOL` cell shadows. No single row order fixes both columns — a shared order means some column's default cell precedes its explicit one, so **every cell must guard on its own row**.

### Options

**A — guard every cell; keep the single matrix.** `cell = out_exp * block` unchanged. `totalize_column` prepends the row's match to each synthesized cell (mirrors `partialbind`'s case guard: top-level case → one `MatchE`; sub-variant case → `SubE` + downcast + `MatchE`):

```ocaml
let guard_of_nottyp ctx exp_in typ_param frees nottyp typ_home =
  let mixop = Mixfix.to_mixop nottyp.it in
  if Il.Eq.eq_typ typ_home typ_param then
    (frees, [ Il.IfPr (Il.MatchE (exp_in, Il.CaseP mixop) $$ (at, Il.BoolT)) $ at ])
  else
    let sub = Il.IfPr (Il.SubE (exp_in, typ_home) $$ (at, Il.BoolT)) $ at in
    let frees, exp_h = Fresh.exp_from_typ ~dim:false ctx.menv frees typ_home in
    let dc  = Il.LetPr (exp_h, Il.DownCastE (typ_home, exp_in) $$ (at, typ_home.it)) $ at in
    let mat = Il.IfPr (Il.MatchE (exp_h, Il.CaseP mixop) $$ (at, Il.BoolT)) $ at in
    (frees, [ sub; dc; mat ])
```

`struct_tablegroup` then builds every cell block through the same `antiunify_clauses` path the singleton table uses, so the one group `param` binds every clause's input var (explicit and synthesized) uniformly:

```ocaml
let clauses_flat =
  List.concat_map (List.map (fun (_s,args,out,prems) -> (args,out,prems) $ at)) covering in
let args_input, paths, _ = Antiunify.antiunify_clauses clauses_flat None in
let param_sl = struct_params_from_args ctx [ param ] args_input |> List.hd in
(* reshape `paths` back into header × cell *)
```

`invoke_table_group` is unchanged. `out_exp` still renders; `block` executes. Cost: `totalize_column` gains the guard; `struct_tablegroup` swaps its per-row `populate` for one `antiunify_clauses`.

**B — execute per column, matrix for render only.** Store each column as its own executable table (its original rows through the proven `struct_table_dec_def`), and keep the refined matrix as a render-only view of output values. `$g.col(x)` runs column `col`'s table directly. Correct by construction (closer stays a catch-all), but splits one artifact into an exec table + a render matrix.

### Open question

A keeps a single matrix where each cell both executes and renders, at the cost of synthesizing per-cell guards in structuring. B reuses the singleton executor unchanged but introduces an exec/render split in the SL representation. Currently leaning A (no exec/render dichotomy). Preferences on which model the SL `tablegroup` should carry?

Reproduction and the harness are local; happy to paste the full `struct` dump of the buggy matrix.

PR #265, opened 2026-08-01
Render PL fallthrough as explicit control-flow labels and cross-group edges

jaehyun1ee:
## Summary

PL prose renders each relation and function as its backtracking control flow, but the flow was opaque: markers `else #N-x` / `fail #N` on a node named `TryI`, cross-group fallthrough drawn as disconnected blocks, and the partiality a nested call adds left unmarked. This branch makes the flow explicit and readable, and derives it in one place — the AST — read by both the annotate golden and the spliced release doc.

Fallthrough now renders as an inline label on each partial premise: `[-> b]` (next arm), `[-> <group>]` (next rulegroup), `[-> ⋅]` (the def's elseblock), `[FAIL]` (no alternative left). The `Try #N:` head becomes `Block:`.

## Commits

1. **Rename `TryI` → `BlockI`** across `Pl.Ast` / `Ll.Ast` / `interp-pl` / `annotate` / splicer (`eval_try_*` → `eval_block_*`). The node drives the `interp-pl` execution tower, not just rendering; semantics and golden are byte-identical.
2. **Encode PL fallthrough in the AST and render it from there.** A structural edge per partial instruction, stamped by an `annotate` pass and read by render + the rulegroup splicer; the render-time derivation (`backtrack.ml`) is deleted.
3. **Style block arms as boxes** — `div.bk-arm` in `docinfo.html`, `role: bk-arm` in the PDF theme, across the p4 / sl / al docs.
4. **Drop the render-time else-anchor trim; align the stamp else-gate** so a def's elseblock always anchors its `[-> ⋅]` label, and an empty elseblock falls to `[FAIL]` rather than a dangling link.
5. **Document the render pass** — a per-function comment carrying a concrete rendered example on every renderer.

## Prose changes

**Cross-group edges — were invisible.** A relation's rulegroups are ordered alternatives, but each was rendered standalone, so a group's last-arm failure dead-ended. The edge now targets the next group using the dispatch tree's own anchor. `Type_alpha`'s `typedefTypeIR` last arm: `[fail #2]` → `[-> newTypeIR]` linking `#Type_alpha-newTypeIR`, the same anchor as its `goto xref:Type_alpha-newTypeIR` dispatch line. It renders in both the annotate golden and the spliced release doc.

**Nested-call partiality.** A call can fail (`invoke_func` raises `Unmatch`), so `Return $foo(x)` backtracks to the enclosing block's next arm. Such `ResultI` / `ReturnI` / `DestructI` now carry the label; a call-free `Return x` stays bare.

nested-call labels

**Dispatch-tree list rendering (pre-existing bug since #261).** The `<Rel> dispatch:` header had no blank line before its `goto` list, so Asciidoctor folded the list into the paragraph — one run-on line with literal `.`/`..` markers. `render_defined_rel_def_dispatch` now emits the blank line.

dispatch tree before/after

**Inline short `goto`.** A guard whose whole body is a single dispatcher `GroupI` folds onto one line (`. If let X be Y: goto Z`), like the existing short `Return`/`Result` fold; multi-step dispatch (nested `If`/`Else`) stays expanded.

inline goto

**Arm bounding boxes (HTML + PDF).** Each block arm's body is wrapped in a `[.bk-arm]` AsciiDoc open block — styled `div.bk-arm` in `docinfo.html` and `role: bk-arm` in the PDF theme. The wrapper attaches via list continuation, so the arm letters/anchors the `[-> b]` links depend on are unchanged.

## How fallthrough is computed

`Pl.Ast` carries a structural edge per partial instruction (no render strings):

```ocaml
and fallthrough = FallGroup of id | FallNext | FallElse | FallFail
```

An `annotate` `Stamp` pass threads the enclosing terminal down the block tree and stamps each partial instruction. Partiality is decided by shared `Pl.Partial` predicates:

```ocaml
let is_partial_instr (instr : instr) : bool =
  match instr.node.it with
  | IfI (exp_cond, _, _, _) -> is_partial_exp exp_cond
  | HoldI _ -> true
  | CaseI (exp, cases, _) -> is_partial_exp exp || List.exists is_partial_case cases
  | GroupI _ | BlockI _ -> false
  | LetI (_, exp_r, _) -> is_partial_exp exp_r
  | RuleI (_, notexp, _, _) -> notexp |> Mixfix.args |> List.exists is_partial_exp
  | ResultI (_, exps) -> List.exists is_partial_exp exps
  | ReturnI exp -> is_partial_exp exp
  | DebugI exp -> is_partial_exp exp
  | DestructI (_, exp) -> is_partial_exp exp
  | CheckLetSubI _ | CheckLetMatchI _ | OptionGetI _ -> true
```

Cross-group edges follow the flattened group sequence (`Collect.collect_groups`, since a relation's groups nest inside the `If`/`Case`/`Hold` dispatch tree rather than as flat sibling arms), and only a def or group body's sole `BlockI` inherits the ambient terminal — a nested `BlockI` fails locally. Render then resolves the structural edge to a label via `Pl.Render.Fallthrough`, which owns the anchor scheme (`anchor_of_group`, `anchor_of_else`) shared with the rulegroup splicer.

This replaces the render-time derivation: `Backtrack.update`/`target`, `terminals_of_groups`, `block_has_fallthrough`, and render's own `exp_has_call` are deleted. One source of truth, read by both prose producers; the AST holds only structural edges, and anchors/letters are resolved in render.

## Validation

- `make build` clean; `make test-speclang` green. The AST-encoding step (commit 2) is **byte-identical** to the prior render-time output — the pass reproduces the exact edges render derived. Marker count `else #` / `fail #` / `Try #` → 0.
- Commit 4 is the only intended golden change: dropping the anchor trim emits a def's else-anchor uniformly (+24 `<span id="…-else">`), and gating both the stamp and the render on a non-empty elseblock means a `Some []` def falls to `[FAIL]` instead of a dangling `[-> ⋅]`. No label flips — the spec has no empty-else defs.
- `make test-run-pl`, `make test-sim-pl` — 0 real failures; the interpreter ignores the new AST field.
- Each prose change validated as rendered HTML with `asciidoctor` (screenshots above; label glyph is now `[-> ]`).

## Outstanding

- Regenerate `docs/p4/P4-16-spec-release.{html,pdf}` in a gem environment on this branch (the dev shell has no `asciidoctor`/`asciidoctor-pdf`) for the combined visual check of boxes, cross-group links, dispatch tree, and inline gotos.


p4spec/lib/lang/pl/render/backtrack.ml
```diff
@@ -34,9 +34,10 @@ end = struct
     { id; display = F.asprintf "#%d" name }
 end
 
-(* Where a failed instruction backtracks to: the next arm, or out of the block *)
+(* Where a failed block resolves: the next arm, or the block's terminal continuation *)
 
-type target = NextArm of string | OutOfBlock
+type terminal = Fail | Goto of { anchor : string; display : string }
```
jaehyun1ee: Instead, define `type goto = { anchor : string; display : string }`, then `type terminal = Fail | Goto of goto`.

p4spec/lib/lang/pl/render/backtrack.ml
```diff
@@ -34,9 +34,10 @@ end = struct
...
+type target = NextArm of string | Terminal of terminal
```
jaehyun1ee: Rename `NextArm` to `Next`.

p4spec/lib/lang/pl/render/render.ml
```diff
@@ -174,6 +176,23 @@ let string_of_cmpop (cmpop : cmpop) : string =
...
+  | UnE (_, _, e) | UpCastE (_, e) | DownCastE (_, e) | SubE (e, _)
+  | MatchE (e, _) | LenE e | DotE (e, _) | IterE (e, _) -> exp_has_call e
+  | BinE (_, _, e1, e2) | CmpE (_, _, e1, e2) | ConsE (e1, e2) | CatE (e1, e2)
+  | MemE (e1, e2) | IdxE (e1, e2) -> exp_has_call e1 || exp_has_call e2
+  | SliceE (e1, e2, e3) -> exp_has_call e1 || exp_has_call e2 || exp_has_call e3
+  | TupleE es | ListE es -> List.exists exp_has_call es
+  | OptE e_opt -> Option.fold ~none:false ~some:exp_has_call e_opt
+  | CaseE notexp -> List.exists exp_has_call (Mixfix.args notexp)
+  | StrE fields -> List.exists (fun (_, e) -> exp_has_call e) fields
+  | UpdE (e_b, _path, e_f) -> exp_has_call e_b || exp_has_call e_f
```
jaehyun1ee: Please respect the variable naming convention. TAL other parts of the file, or the interpreter code. `e` should be `exp`. `MemE (e1, e2)` should be `MemE (exp_e, exp_s)`, and so on...

p4spec/lib/lang/pl/render/render.ml
```diff
@@ -1089,36 +1135,56 @@ and render_group_instr ~(level : int) ~(backtrack : Backtrack.ctx option)
...
+        Adoc.(text "{empty}" ++ prose_anchor)
+    in
+    let body =
+      Adoc.seq_block
+        (List.map
+           (render_instr ~level:level_body ~backtrack:(Some backtrack)
+              ~dispatcher)
+           arm)
+    in
+    (* Box the arm body; list continuation keeps the letter on `head` *)
```
jaehyun1ee: Delete the comment.

p4spec/lib/lang/pl/render/render.ml
```diff
@@ -1446,33 +1533,79 @@ let render_rulegroup (hints : Annot.hints) (_id_rulegroup : id) (id_rel : id)
...
+        render_instrs ~backtrack:(Some backtrack) block
+  in
+  title ^ ":\n" ^ Adoc.ser_block body
 
 let render_defined_rel_def_dispatch
     ((id_rel, _rel_signature, _exps, block, _elseblock_opt) : rel) : string =
-  let head =
-    Adoc.inline_block Adoc.(text (string_of_relid id_rel) ++ text " dispatch:")
+  (* the "... dispatch:" header is a paragraph; without a blank line before the
+     list Asciidoctor folds the list into the paragraph as literal text *)
```
jaehyun1ee: Remove the comment.

p4spec/lib/lang/pl/render/backtrack.ml
jaehyun1ee: Extracted `type goto = { anchor : string; display : string }`; `terminal = Fail | Goto of goto`. 9d40c84b

p4spec/lib/lang/pl/render/backtrack.ml
jaehyun1ee: Renamed `NextArm` → `Next`. 9d40c84b

p4spec/lib/lang/pl/render/render.ml
jaehyun1ee: Named sub-expressions per the constructor convention (`exp_l`/`exp_r`, `exp_e`/`exp_s`, `exp_b`/`exp_i`, …). 9d40c84b

p4spec/lib/lang/pl/render/render.ml
jaehyun1ee: Deleted. 9d40c84b

p4spec/lib/lang/pl/render/render.ml
jaehyun1ee: Deleted. 9d40c84b

Issue #175, opened 2026-07-08
Plan: port structured diagnostics and LSP support from SpecTecX


KunJeong:
SpecTecX (kaist-plrg/spectecx) has grown structured diagnostics and an LSP server. This issue plans porting these to P4-SpecTec, and lays out the required steps so that we can estimate the work and split if necessary.

The port is relatively cheap because the core parse and elaboration logic are still almost unchanged: the error sites in `lib/pass/elaborate/elab.ml` are identical to SpecTecX, and the parser (10) and lexer (16) sites correspond one to one. So the error messages and contents can be ported one-on-one. In addition, SpecTecX adds a diagnostic record, a renderer, a bridge from backtracking failure traces to the record, and a small LSP server. Estimated total: about two weeks, mostly mechanical.

The one gap is that `elab_spec` raises at the first error, so collecting several diagnostics per run should be added before anything else (Phase 1 below). Everything else needed is already present.

## Source pull requests ([SpecTecX](https://github.com/kaist-plrg/spectecx))

| PR | Contents |
|---|---|
| kaist-plrg/spectecx#27) | Diagnostic record and ANSI renderer; per-pass error modules; `with_diagnostics` facade |
| kaist-plrg/spectecx#43) | Elaboration diagnostics: per-site `code` catalog and failure-trace bridge |
| kaist-plrg/spectecx#52) | Parse and lexer diagnostics |
| kaist-plrg/spectecx#55) | Rendering refinements |
| kaist-plrg/spectecx#59) | Interpreter failure traces: the `of_failtraces` bridge |
| kaist-plrg/spectecx#82) | LSP server over `linol-eio`; editor configuration |
| kaist-plrg/spectecx#85) | Whole-spec diagnosis in the LSP; shared spec-file collection; prevents LSP crash |

kaist-plrg/spectecx#55 refinements will be collapsed into kaist-plrg/spectecx#27.

## Plan

Ordered by dependency, following the order SpecTecX built the feature in.

### Phase 1: groundwork (~= kaist-plrg/spectecx#27)

Port the `diag/` library (`record`, `render`, `ansi`, `source_cache` plus their .mli counterparts). Adapt the failure-trace bridge (`of_failtraces`, kaist-plrg/spectecx#59) to this repository's `Failtrace of region * string * failtrace list`. Add the collection logic (`with_diagnostics`, backed by a `Sink` that accumulates diagnostics instead of printing them).

[...]

PR #277, opened 2026-08-13
Refactor exception-based error handling to typed results through a single facade

KunJeong:
## Motivation

Failure handling in the pipeline and runners was done through exceptions, and the exception types lived in a shared `Util.Error` module that every fallible module pulled in through `open`. This led to two problems:
- Signatures were dishonest. A function that ran a pass or initialized a runner returned a plain value but could fail with an exception, so every CLI-facing caller wrapped the call in a `try` and re-listed every exception that might happen.
- Ownership was enforced by convention only. Any module could raise or catch any error, and nothing restricted which failures a given pass could produce.

This is phase 1 of the diagnostics/LSP port (#175): make every failure flow as a typed value through a single boundary, so following work can attach positions and render diagnostics in one place instead of at each `try`.

### 1. A pass returns a result instead of raising an exception

The pipeline stages returned bare values and raised exceptions on failure. They now return a result over `Pass.error`, which is made abstract by a new `pass.mli`.

```ocaml
(* before: each command re-lists the pass exceptions it might see *)
fun () ->
  try
    let spec_sl = Pass.structure ~final:true paths_spec in
    Format.printf "%s\n" (Sl.Print.string_of_spec spec_sl)
  with
  | ParseError (at, msg) | ElabError (at, msg) | StructError (at, msg) ->
      Format.printf "%s\n" (string_of_error at msg)
```
```ocaml
(* after: pass/structure/structure.ml catches its exception and returns a result *)
let struct_spec ~(final : bool) (spec : Al.spec) : (Sl.spec, error) result =
  try Ok (Struct.struct_spec ~final spec)
  with Error.StructError (at, msg) -> Error { at; msg }

(* pass.ml converts each pass's error into Pass.error *)
let struct_spec ~final spec_al =
  Structure.struct_spec ~final spec_al
  |> Result.map_error (fun e -> StructError e)

(* pass.mli makes the error type abstract *)
type error
val structure : final:bool -> string list -> (Lang.Sl.spec, error) result
```

### 2. Each module owns its error

`Util.Error` declared twelve exceptions and their raise helpers and was `open`ed by every fallible module, so any module could raise or catch any failure. It now keeps only error rendering and warning printing. Each pass declares its own exception and converts it to its own error type at its entry point, and `Pass.error` is the union.

```ocaml
(* before: util/error.ml owns all errors *)
exception ElabError of region * string
exception StructError of region * string
(* ... *)
let error_elab at msg = raise (ElabError (at, msg))
```

```ocaml
(* after: pass/elaborate/error.ml owns the elaboration failure *)
exception ElabError of region * string
let error at msg = raise (ElabError (at, msg))

(* pass.ml: unions submodules' error types *)
type error =
  | ParseError of Frontend.Parse.error
  | ElabError of Elaborate.error
  | AlgoError of Algo.error
  | StructError of Structure.error
  | ProseError of Annotate.error
```
### 3. Callers compose through one facade

The new `p4spectec.ml(i)` acts as a facade, exposing pass/runner/command operations as functions whose failures share a single `Error.t`, which is a union of all pass, runner, and command errors. Callers chain with `let*`, match once, and render with `Error.to_string`. Effectively, `Pass`, `Backend_sim`, and `Backend_boot.Build` no longer appear in any consumer under `bin` and `test`.
```ocaml
(* before: the builder raises, and each caller re-invents a CommandError *)
let spec_sim, (module Simulator) =
  Backend_sim.Build.build ?arch ~final:true mode paths_spec
in
let spec_sl =
  match spec_sim with
  | SL spec_sl -> spec_sl
  | _ -> raise (CommandError "instruction coverage is only supported for SL")
in
```
```ocaml
(* after: every step is a result threaded through the facade *)
let* spec_sim = P4spectec.spec_of_mode mode paths_spec in
let* simulator = P4spectec.build_sim ?arch spec_sim in
let (module Simulator : SIM) = simulator in
let* spec_sl =
  match spec_sim with
  | SL spec_sl -> Ok spec_sl
  | _ -> Error (Error.CommandError "instruction coverage is only supported for SL")
in

(* error.ml: the one type every entry point reports, and its renderer *)
type t =
  | PassError of Pass.error
  | RunError of Run.error
  | CommandError of string

let to_string (e : t) : string =
  let at, msg = to_region_msg e in
  Util.Error.string_of_error at msg
```
## Scope

No spec output changes, and the only behavioral differences are the minor fixes below.

### Minor Fixes

Several bugs are found and fixed along the way, most of them converting uncaught exceptions to graceful failures.
- **Parser:** path resolution (`expand_path`) moves into `Frontend.Parse, and `Sys_error` is caught by `parse_files` instead of escaping uncaught.
- **Tower:** `P4spectec.tower_of_file` now catches `Sys_error`, `Yojson.Json_error`, and `Yojson.Basic.Util.Type_error`, and returns it as a `CommandError`.
- **AL Interpreter:** an `ExternError` escaped uncaught and was caught later by boot/main. `Interp-al` now catches it internally, matching the SL and PL interpreters, so it is complete without relying on callers.
- **Dangling Coverage:** the `cover_run_dangling` command used to report "instruction coverage is only supported for SL", it now says "dangling coverage is only supported for SL".

## Future Work

Attaching source positions to `Error.t` and rendering real diagnostics is the next phase of #175.

## Commit Log

- refactor(build): build the tower and simulator from a prebuilt spec
- refactor(pass): return results from the pipeline aggregators
- refactor(pass): give each pass its own error type
- refactor(error): move each exception to the module that owns it
- refactor(error): give callers a single error type

### DIFF
--- a/p4spec/lib/backend-boot/build.ml
+++ b/p4spec/lib/backend-boot/build.ml
@@ -1,5 +1,5 @@
 module Run = Runtime.Dynamic_Runner.Signature
-open Error
+open Util.Source
 
 let ( let* ) = Result.bind
 
@@ -30,10 +30,50 @@ let spec_of_mode (mode : Run.mode) (paths_spec : string list) :
       Ok (PL spec_pl : Run.spec)
   | Empty_mode -> assert false
 
+(* The specs of a tower, paired with the level each one belongs to, ordered from
+   the target down to the boot level *)
+
+type tower_specs = {
+  target : Config.level * Run.spec;
+  interms : (Config.level * Run.spec) list;
+  boot : Config.level * Run.spec;
+}
+
+let specs_of_tower (tower : Config.tower) : (tower_specs, Pass.error) result =
+  let* spec_target = structured_spec [ tower.level_target.layer.specdir ] in
+  let* specs_interm =
+    tower.levels_interm |> List.rev
+    |> map_result (fun (level : Config.level) ->
+           let* spec = structured_spec [ level.layer.specdir ] in
+           Ok (level, spec))
+  in
+  let* spec_boot = spec_of_mode tower.mode [ tower.level_boot.layer.specdir ] in
+  Ok
+    {
+      target = (tower.level_target, spec_target);
+      interms = specs_interm;
+      boot = (tower.level_boot, spec_boot);
+    }
+
+(* The SpecTec interface backing a non-target level *)
+
+let spectec_interface_of (interface : Config.interface) :
+    ((module Spectec.INTERFACE_SPECTEC), Run.error) result =
+  match interface with
+  | P4_interface ->
+      Error
+        {
+          Run.at = no_region;
+          msg = "P4 interface not supported outside of target level";
+        }
+  | AL_interface -> Ok (module Interface.SpecTec_AL : Spectec.INTERFACE_SPECTEC)
+  | SL_interface -> Ok (module Interface.SpecTec_SL : Spectec.INTERFACE_SPECTEC)
+
 (* Building a tower *)
 
 let build_target ?(cache = true) ?(det = false) ?(guard = false)
-    (level : Config.level) (spec : Run.spec) =
+    (level : Config.level) (spec : Run.spec) :
+    ((module Run.RUNNER), Run.error) result =
   (* Create the target runner *)
   let (module Runner_target) =
     match level.interface with
@@ -55,19 +95,15 @@ let build_target ?(cache = true) ?(det = false) ?(guard = false)
                   (Interp_sl.Interp.Make)
                   (Interp_pl.Interp.Make) : Run.RUNNER)
   in
-  Runner_target.init ~cache ~det ~guard spec;
-  (module Runner_target : Run.RUNNER)
+  let* () = Runner_target.init ~cache ~det ~guard spec in
+  Ok (module Runner_target : Run.RUNNER)
 
 let build_interm ?(cache = true) ?(det = false) ?(guard = false)
     (module Runner_above : Run.RUNNER) (level : Config.level) (spec : Run.spec)
-    =
+    : ((module Run.RUNNER), Run.error) result =
   (* Create the intermediate runner *)
-  let (module Interface_SpecTec) =
-    match level.interface with
-    | P4_interface ->
-        error_no_region "P4 interface not supported outside of target level"
-    | AL_interface -> (module Interface.SpecTec_AL : Spectec.INTERFACE_SPECTEC)
-    | SL_interface -> (module Interface.SpecTec_SL : Spectec.INTERFACE_SPECTEC)
+  let* (module Interface_SpecTec : Spectec.INTERFACE_SPECTEC) =
+    spectec_interface_of level.interface
   in
   let (module Runner) =
     (module Runner.Make.Make_nonrec
@@ -77,19 +113,15 @@ let build_interm ?(cache = true) ?(det = false) ?(guard = false)
               (Interp_sl.Interp.Make)
               (Interp_pl.Interp.Make) : Run.RUNNER)
   in
-  Runner.init ~cache ~det ~guard spec;
-  (module Runner : Run.RUNNER)
+  let* () = Runner.init ~cache ~det ~guard spec in
+  Ok (module Runner : Run.RUNNER)
 
 let build_boot ?(cache = true) ?(det = false) ?(guard = false)
     (module Runner_above : Run.RUNNER) (level : Config.level) (spec : Run.spec)
-    =
+    : ((module Run.RUNNER), Run.error) result =
   (* Create the booter *)
-  let (module Interface_SpecTec) =
-    match level.interface with
-    | P4_interface ->
-        error_no_region "P4 interface not supported outside of target level"
-    | AL_interface -> (module Interface.SpecTec_AL : Spectec.INTERFACE_SPECTEC)
-    | SL_interface -> (module Interface.SpecTec_SL : Spectec.INTERFACE_SPECTEC)
+  let* (module Interface_SpecTec : Spectec.INTERFACE_SPECTEC) =
+    spectec_interface_of level.interface
   in
   let (module Booter) =
     (module Runner.Make.Make_nonrec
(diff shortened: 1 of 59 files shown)


p4spec/lib/p4spectec.ml
```diff
@@ -0,0 +1,87 @@
...
+  let* (specs : Boot_build.tower_specs) =
+    Boot_build.specs_of_tower tower
+    |> Result.map_error (fun e -> Error.PassError e)
+  in
+  let* booter =
+    Boot_build.build_tower ~cache ~det ~guard specs
+    |> Result.map_error (fun e -> Error.RunError e)
+  in
+  let _, spec_boot = specs.boot in
+  Ok (spec_boot, booter)
```
jaehyun1ee: Reorder this, after `tower_of_file`?

p4spec/lib/p4spectec.mli
```diff
@@ -0,0 +1,76 @@
...
+  ?det:bool ->
+  ?guard:bool ->
+  Backend_boot.Config.interface ->
+  Runtime.Dynamic_Runner.Signature.spec ->
+  (module Runtime.Dynamic_Runner.Signature.RUNNER) result
+
+val tower_of_file :
+  string -> Backend_boot.Config.target -> Backend_boot.Config.tower result
+
+(* Builds every level of the tower, returning the boot spec and its runner *)
```
jaehyun1ee: Delete the comment?

p4spec/test/test_common.ml
```diff
@@ -1,8 +1,19 @@
...
+let ( let* ) = Result.bind
+
+let build_sim ?cache ?det ?guard ?(arch : string option) mode paths_spec =
+  match
+    let* spec_sim = P4spectec.spec_of_mode mode paths_spec in
+    let* simulator = P4spectec.build_sim ?cache ?det ?guard ?arch spec_sim in
+    Ok (spec_sim, simulator)
+  with
+  | Ok (spec_sim, simulator) -> (spec_sim, simulator)
+  | Error e -> failwith (Error.to_string e)
```
jaehyun1ee, line 16: Instead of having an aux function `build_sim`, can we do the pattern match inside the callers, likewise `main.ml`?

p4spec/lib/frontend/parse.ml
```diff
@@ -40,17 +46,31 @@ let parse_mixop str =
...
-  let ic = open_in file in
   try
+    let ic = open_in file in
     Fun.protect
       (fun () -> with_lexbuf file (Lexing.from_channel ic) Parser.spec)
       ~finally:(fun () -> close_in ic)
   with Sys_error msg ->
     error (Source.region_of_file file) ("i/o error: " ^ msg)
 
+(* A directory expands to its .watsup files. A plain path stands for itself. *)
```
jaehyun1ee: I prefer deleting this comment.

p4spec/lib/backend-boot/build.ml
```diff
@@ -79,64 +131,33 @@ let build_boot ?(cache = true) ?(det = false) ?(guard = false)
...
-        let spec_al = Pass.algo [ level.layer.specdir ] in
-        (AL spec_al : Run.spec)
-    | SL_mode ->
-        let spec_sl = Pass.structure ~final:true [ level.layer.specdir ] in
-        (SL spec_sl : Run.spec)
-    | PL_mode ->
-        let spec_pl = Pass.annotate [ level.layer.specdir ] in
-        (PL spec_pl : Run.spec)
-    | Empty_mode -> assert false
-  in
```
jaehyun1ee, line 95: Not entirely sure why this should be deleted.

p4spec/lib/backend-boot/build.ml
```diff
@@ -146,18 +167,5 @@ let build_null ?(cache = true) ?(det = false) ?(guard = false) (mode : Run.mode)
...
-    | SL_mode ->
-        let spec_sl = Pass.structure ~final:true paths_spec in
-        (SL spec_sl : Run.spec)
-    | PL_mode ->
-        let spec_pl = Pass.annotate paths_spec in
-        (PL spec_pl : Run.spec)
-    | Empty_mode -> assert false
-  in
-  Runner.init ~cache ~det ~guard spec;
-  (spec, (module Runner : Run.RUNNER))
```
jaehyun1ee, line 163: And this as well?

p4spec/test/test_common.ml
jaehyun1ee, line 16: Given that we have `(let *)`, can't we delete this wrapper, like in `main.ml`?

jaehyun1ee (changes requested):
Thanks! Happy to see better management of errors. I have some minor comments.

KunJeong:
Thanks for the quick review! I've applied the stylistic fixes, and I'll clarify below on the two remaining comments.

### 1. The mode -> spec block

The logic isn't dropped, it's moved to [`Backend_boot.Build.spec_of_mode`](https://github.com/kaist-plrg/p4-spectec/pull/277/changes#diff-02001f0025ee6cac13c079927236e710f960fe21e7fe617984328312be1854b7R15-R27). The commit message of 94ed30c explains this at a high level, but I probably should've explained it better, because the mechanism isn't apparent from the diff.

Before, `build_boot` took a `mode`, built the spec, and returned it back out. `build_tower` had to store it via a ref `spec_boot`. Now, `specs_of_towers` builds each layer's spec first via `spec_of_mode`, which is then passed via  to `build_boot` via its new spec parameter.

This gives us two benefits:
- It eliminates the `spec_boot`'s ref-storing workaround.
- It also splits the function at a clear semantic boundary. `spec_of_mode`/`specs_of_tower` build spec(s) from filepaths, so their failures are `Pass.error`s. The builder functions, on the other hand, return a `Run.error`. Splitting the two functions lets them have return types that accurately describe how they may fail.

### 2. `Test_common.build_sim`

This is actually a deliberate decision to keep a narrow scope. `build_sim` has seven callers, three in `test_common.ml` and four in other test runners. All test runners are currently exception-driven, including `build_sim`. So inlining/converting to `let*` at the callers would mean using result types and `let*` for `build_sim` only, repeated at seven call sites. The current choice is to keep the `test/` changes as minimal as possible and avoid any behavior changes by converting `build_sim`'s result type back to an exception. Converting all of the test harness to use result types might be desirable later for maximum consistency, but I feel it's best not to include in this PR as it needs its own design and discussion.

I know it may look inconsistent, but it's actually coherent in that we're focusing on `lib/` for the refactor, and for stuff in `bin/` and `test/` the goal is to keep all observable behavior unchanged.

jaehyun1ee:
Thanks for the explanation.

> The current choice is to keep the test/ changes as minimal as possible and avoid any behavior changes by converting build_sim's result type back to an exception.

Merging, but we should revisit this though, for consistency.

jaehyun1ee (approved):

PR #287, opened 2026-08-21
fix: totalize bitstr decode at width 0 and wrap unsigned negation

jihongmin63:
Two independent defects at the boundary of fixed-width arithmetic.

$bitstr_to_int(w, n) normalizes n into [-2^(w-1), 2^(w-1)) by adding or subtracting 2^w until the value lands inside. At w = 0 that interval is empty, so the base case is unreachable and n = 0 oscillates 0 -> -1 -> 0 forever. bit<0> reaches it because the unsigned clauses of + - * / % and << >> decode as well as the signed ones, so `const bit<0> B = A + A;` makes the checker spin. Guard the recursion at w <= 0, where bit<0>'s single value 0 is the answer.

$un_minus's unsigned clause computes 2^w - i without reducing modulo 2^w, so negating 0 yields 2^w, outside the range of bit<w>. Re-encode through $int_to_bitstr the way the signed clause already does.

It follows previous [PR](https://github.com/kaist-plrg/p4-spectec/pull/284).

### DIFF
--- a/p4spec/lib/interface/builtin/numerics.ml
+++ b/p4spec/lib/interface/builtin/numerics.ml
@@ -103,11 +103,13 @@ let pow2 (add : value -> unit) (at : region) (targs : targ list)
 (* dec $bitstr_to_int(int, bitstr) : int *)
 
 let rec bitstr_to_int' (w : Bigint.t) (n : Bigint.t) : Bigint.t =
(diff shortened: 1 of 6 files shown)


jaehyun1ee:
Thanks!

PR #292, opened 2026-09-04
spec: guard $strip_prefix_rec/$strip_suffix_rec against the empty pattern

jihongmin63:
# spec: guard `$strip_prefix_rec`/`$strip_suffix_rec` against the empty pattern

Both helpers in `spec/0-aux/0.0-stdlib.watsup` diverge when the pattern is `""`.

## The loop

`$starts_with(t, "")` is always `true` — `|""| <= |t|` and `t[0 : 0] = ""` hold for
every `t`. `$strip_prefix(t, "")` is the identity (`texts.ml:47` returns
`String.sub t 0 |t|`). So the recursive clause of `$strip_prefix_rec(t, "")` fires
with unchanged arguments forever, while the base clause, guarded by
`~$starts_with(t, "")`, can never fire. `$ends_with`/`$strip_suffix` give the same
loop for `$strip_suffix_rec`.

The implicit termination measure — `|t|` shrinks by `|t_prefix|` per step — assumes
`|t_prefix| >= 1`, and nothing established that. Same family as the width-0 fix in
#287: an inductive step whose measure degenerates at the boundary.

## Fix

Two edits per helper:

- `-- if $(|t_prefix| > 0)` on the recursive clause, making the measure's assumption
  a premise. Arithmetic rather than `t_prefix =/= ""`, since `=/=` routes through
  generic structural equality and bloats the downstream analysis slice.
- base clause guard `-- if ~$starts_with(t, t_prefix)` → `-- otherwise`. Keeping the
  old guard would make the function *partial* at `""` — recursive clause blocked by
  the new premise, base clause blocked by its own — trading the loop for a stuck
  call. With `otherwise`, `$strip_prefix_rec(t, "")` = `t`, the sensible fixpoint.

## No behavior change

`$strip_prefix_rec` has no call sites. `$strip_suffix_rec` has three — all in
`spec/9-arch/9.1-table-interface.watsup` (204, 232, 286), all passing the literal
`"*"`. `|"*"| > 0`, so the new premise holds at every call site and `otherwise` is
reached in exactly the cases the old guard reached. The loop is latent; this came
out of termination analysis, not an observed hang.

## Testing

`make test-fast` passes, and `p4spectec elab` over the full spec succeeds with the
new premise and the `clause -1` fallback as intended.

Four `test/lang` goldens are promoted. `struct` and `annotate` look enormous but
aren't: outside these two functions every changed line is `Dangling#N` counter
renumbering shifted by the added premises — checked line by line, 8 real changes
in each.

`@backend-adoc` already fails on `main`, independently of this change —
`pl.expected` carries three stale hunks there (the `unroll_aliasType` ->
`unroll_aliasTypeIR` rename, the zero-width-slice `ite`), confirmed on a clean
tree. `make promote` would drag those into this PR, so only this change's hunk was
applied to that golden by hand; the suite stays red on the other three.

### DIFF
--- a/p4spec/test/backend-adoc/pl.expected
+++ b/p4spec/test/backend-adoc/pl.expected
@@ -94,17 +94,21 @@ xref:ends_with[The text ``t`` ends with ``t~suffix~``]
 
 xref:strip_prefix_rec[The text ``t`` with the prefix ``t~prefix~`` removed many times as possible]
 
-. If xref:starts_with[the text ``t`` starts with ``t~prefix~``]:+++<sub class="bk-mark">[FAIL]</sub>+++
- .. Let ``t~stripped~`` be xref:strip_prefix[the text ``t`` with the prefix ``t~prefix~`` removed].+++<sub class="bk-mark">[FAIL]</sub>+++
- .. Return xref:strip_prefix_rec[the text ``t~stripped~`` with the prefix ``t~prefix~`` removed many times as possible].+++<sub class="bk-mark">[FAIL]</sub>+++
-. Else: return ``t``.
+. Check that the length of ``t~prefix~`` is greater than ``0``.
+. Check that xref:starts_with[the text ``t`` starts with ``t~prefix~``].+++<sub class="bk-mark">[<a href="#strip_prefix_rec-else">→ ⋅</a>]</sub>+++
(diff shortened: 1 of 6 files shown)


jaehyun1ee:
I would much appreciate if the PR description was more polished ... without wrap-arounds in text, no redundant title in the description when there is already a title for the PR, and so on. Aside from the content of the code edits, the presentation matter for making contributions.

For example, I'm not entirely sure if you have reviewed the last paragraph. This reads more like a monologue of an agent, not sure what it implies for the reviewer.

PR #297, opened 2026-09-08
Add structured diagnostics across P4-SpecTec

KunJeong:
## Motivation

P4-SpecTec failures carried little more than a source region and message. This discarded stable error codes, explanations, related locations, and structured backtracking context before callers could render or inspect them.

Warnings also printed immediately, while command, parser, and interpreter boundaries frequently reduced failures to strings. These limitations prevented consistent CLI reports and left no suitable diagnostic interface for later LSP work.

This PR implements the structured diagnostic groundwork and per-pass emission described by Phases 1 and 2 of #175, including the interpreter work in Phase 2.3. Multi-error elaboration recovery and the Phase 3 LSP server remain separate.

## Summary of Changes

- Add `Diagnostic.t` with severity, source region, stable code, message, detail, related locations, trace, and producer source.
- Add diagnostic collection, source-annotated rendering, source caching, and structured failtrace conversion.
- Render CLI diagnostics on stderr while preserving successful command output on stdout.
- Construct diagnostics at parser, elaborator, algorithm, prose annotation, command, runtime, and interpreter boundaries.
- Replace unreachable defensive errors with assertions whose comments state the producer invariant.
- Preserve complete parser diagnostics and nested failtraces through interpreter, extern, boot, simulator, and CLI boundaries.
- Wrap Petr4's region-and-message parse errors once as uncoded `source:"p4"` diagnostics.
- Route active warnings through the diagnostic collector.
- Give each of the 124 reachable authored diagnostic sites one stable code and one corresponding snapshot test.

## Provenance and Local Work

The shared diagnostic representation, collector, renderer, pass boundary, and failtrace bridge derive from SpecTecX. Ported commits retain `Original-commit` trailers identifying their source commits.

The counts below cover authored diagnostic sites. Each site has one code and one snapshot test. SpecTecX-equivalent means the same underlying check, these can still have local improvements in wording.

| Area | SpecTecX-equivalent | P4-SpecTec-specific | Total |
| --- | ---: | ---: | ---: |
| Parser | 21 | 2 | 23 |
| Elaboration | 51 | 27 | 78 |
| Algorithm | 5 | 10 | 15 |
| Prose annotation | 0 | 3 | 3 |
| Command | 0 | 3 | 3 |
| Boot | 0 | 1 | 1 |
| Simulator | 0 | 1 | 1 |
| **Total** | **77** | **47** | **124** |

<details>
<summary>P4-SpecTec-specific diagnostic sites (47)</summary>

- Parser checks: `parse/malformed-mixop`, `parse/input-path-io-error`
- Extern declarations: `elab/extern-function-redefined`, `elab/extern-dec-tparam-not-distinct`, `elab/extern-relation-redefined`, `elab/extern-relation-rules`, `elab/extern-syn-invalid-id`, `elab/typ-define-extern`, `elab/extend-non-variant-extern`
- Otherwise clauses and rule groups: `elab/function-otherwise-redefined`, `elab/relation-otherwise-redefined`, `elab/rule-group-redefined`, `elab/iter-otherwise-premise`, `elab/prem-multiple-otherwise`, `elab/rule-multiple-otherwise`, `elab/rule-otherwise-with-other-rule`, `elab/rule-relation-mismatch`
- Table functions and rows: `elab/table-function-redefined`, `elab/table-function-undefined`, `elab/table-function-required`, `elab/table-rows-redefined`, `elab/table-function-parameter`, `elab/table-missing-rows`, `elab/table-non-bool-return`
- Additional type and hint checks: `elab/relation-input-hint-out-of-bounds`, `elab/dec-missing-clauses`, `elab/extend-non-variant-tparam`, `elab/typ-missing-definition`, `elab/syn-tparam-not-distinct`
- Otherwise-body analysis: `algo/condition-in-otherwise`, `algo/function-call-in-otherwise`, `algo/relation-call-in-otherwise`
- Table analysis: `algo/table-row-binding-shape`, `algo/pattern-named-variant-expected`, `algo/pattern-extern-type`, `algo/pattern-incomplete`, `algo/pattern-non-variant-type`, `algo/pattern-overlap`, `algo/repeated-table-binding`
- Prose hints: `prose/hint-fields-arity-mismatch`, `prose/hint-fields-text-expected`, `prose/hint-placeholder-out-of-bounds`
- Command and backend checks: `command/boot-source-conflict`, `command/boot-source-required`, `command/splice-file-count-mismatch`, `boot/unsupported-interface`, `sim/unsupported-architecture`

</details>

An important rule of thumb is that **there is exactly one diagnostic test for each diagnostic site.** When a diagnostic site cannot be reached with a test, it is confirmed unreachable and replaced with an explicit assertion, with a one-line comment above that explains why the invariant holds. Each diagnostic site was carefully reviewed one by one to provide informative and clear messages.

Each of the 124 diagnostic tests triggers exactly one site and checks the diagnostic’s displayed output. This includes seven parser tests whose inputs are constructed in OCaml test code to exercise invalid bytes, mixfix parsing, and filesystem errors.

Other tests check formatting, verify that diagnostic fields and nested interpreter failure traces survive passing between components, and check that valid input produces no diagnostic. These tests are outside the count of 124.

The port also improves some SpecTecX diagnostics. For example, a duplicate-parameter error points at the repeated parameter and shows its first declaration. An argument-count error reports both the expected and supplied counts.

## Review Guide

1. Review the diagnostic representation and transport first. `boundary.expected` and `interp.expected` cover parser, interpreter, extern, boot, and simulator boundaries.
2. Review SpecTecX-equivalent diagnostics through the parse and elaboration snapshots. Ported commits include `Original-commit` trailers for comparison.
3. Spend most semantic review time on the 47 P4-SpecTec-specific sites and the assertion comments.

## Diagnostic Conventions

Since there are 124 diagnostic sites, several conventions were established for consistency between the messages. These conventions should be respected when adding new diagnostics.

### Message, Detail, and Labels

- **Name the problem.** Identify the affected construct and say what is wrong with it. Use source-language terms such as `function`, not internal names such as `FuncD`.
- **Make mismatches concrete.** Give the expected and actual counts or types when both are known. Say what each count measures, such as arguments or field names.
- **Use consistent wording.** Use the same terminology and sentence pattern for analogous errors, even when they have different codes.
- **Give detail a separate purpose.** Use it to explain a non-obvious restriction, accepted form, or consequence. Omit it when it would only repeat the message.
- **Explain each related location.** Its label should say why that span matters, such as `first declaration` or `parameter declaring the expected type`.

### Fields and Behavior

- **Severity:** Report an error when the operation fails. Use a warning only when the operation can still complete successfully.
- **Assertions:** Invalid user input must produce a diagnostic. Use an assertion only when an earlier check or construction guarantees the condition, and state that guarantee in the comment.
- **Codes and coverage:** Give each diagnostic site its own stable code and one snapshot test, even if another site has identical wording.
- **Primary location:** Highlight the smallest construct responsible for the error. For missing input, point where it belongs when that position is known.
- **Related locations:** For duplicates, highlight the repeated occurrence and relate the first. When a declaration determines an expected type or arity, relate that declaration.
- **Source:** Identify the component that detected the error. Components forwarding the diagnostic must preserve its source and all other fields.

## Performance

To check for performance degradation, the base revision and the diagnostic implementation were compared using `make test-run-al` and `make test-run-sl`.

| Measurement | Base | This branch | Change |
| --- | ---: | ---: | ---: |
| AL wall time | 170.01 s | 172.48 s | +1.45% |
| SL wall time | 137.16 s | 138.45 s | +0.94% |
| Combined wall time | 307.17 s | 310.93 s | +1.22% |
| Peak RAM usage | 661.7 MB | 659.6 MB | -0.31% |

The performance difference seems negligible.

## Known Limitations

- Error collection and recovery in elaboration is currently missing and deferred.
- LSP support from Phase 3 of #175 is left to the next PR.

## Commit Log

- feat(diagnostic): represent structured diagnostics
- feat(diagnostic): collect emitted warnings
- feat(diagnostic): render source-annotated reports
- feat(p4spectec): collect a run's diagnostics
- feat(cli): render diagnostics to stderr
- feat(elaborate): report warnings as diagnostics
- feat(error): build a diagnostic at each pass failure
- feat(parse): tag parser diagnostics with codes
- feat(elaborate): tag elaboration diagnostics with codes
- feat(elaborate): surface the elaboration failtrace as a trace
- feat(algo): tag binding-analysis diagnostics with codes
- refactor(structure): assert pass input invariants
- feat(annotate): tag prose-generation diagnostics with codes
- feat(interp): preserve structured interpreter failures
- feat(diagnostic): route the remaining warnings through the collector


dune-project
```diff
@@ -23,4 +23,7 @@
     (core (>= "v0.17.0"))
     (core_unix (>= "v0.17.0"))
     (ppx_let (>= "v0.17.0"))
-    (ppx_deriving_yojson (>= "3.9.1"))))
+    (ppx_deriving_yojson (>= "3.9.1"))
+    uucp
+    uuseg
+    uutf))
```
jaehyun1ee, line 29: Should we also update the installation guide in our README and Dockerfiles?

p4spec/lib/frontend/parser.mly
```diff
@@ -225,7 +228,9 @@ nottyp :
     {
       match $1 with
       | NotationT nottyp -> nottyp
-      | _ -> error (at $sloc) "expected notation type"
+      | PlainT plaintyp ->
+          error ~code:Notation_type_expected ~detail:"A notation type includes literal tokens like `|-` or `:` that rules pattern-match against. A bare type like `nat` names a set of values without any tokens, so it cannot serve as a relation body." (at $sloc)
+            (Format.asprintf "relation body must be a notation type, but got plain type %s" (Diagnostic.quote (El.Print.string_of_plaintyp plaintyp)))
```
jaehyun1ee: "[...] serve as a relation body" → "[...] serve as a relation signature"?
Similarly, "relation body must be [...]" → "relation signature must be [...]"?

p4spec/lib/frontend/error.ml
```diff
@@ -1,5 +1,64 @@
...
+  | Unclosed_block_comment -> "unclosed-block-comment"
+  | Notation_type_expected -> "notation-type-expected"
+  | Struct_no_fields -> "struct-no-fields"
+  | Variant_no_cases -> "variant-no-cases"
+  | Syntax_empty_body -> "syntax-empty-body"
+  | Syntax_no_ids -> "syntax-no-ids"
+  | Hint_on_plain_bar_single -> "hint-on-plain-bar-single"
+  | Hint_on_plain_bar_multi -> "hint-on-plain-bar-multi"
+  | Hint_on_plain_no_bar_single -> "hint-on-plain-no-bar-single"
+  | Hint_on_plain_no_bar_multi -> "hint-on-plain-no-bar-multi"
```
jaehyun1ee: I'm not sure it is worth classifying "hint-on-plain{-no}?-bar-{single/multi}". Can we just merge them into one?

p4spec/lib/frontend/parser.mly
```diff
@@ -24,6 +24,9 @@ let positions_to_region position_left position_right =
 let at (position_left, position_right) = positions_to_region position_left position_right
 let (@@@) it pos = it $ at pos
 
+let error_hint_on_plain_type code (hint : El.hint) =
+  error ~code ~detail:"A plain type definition aliases an existing type, as in `syntax x = nat`. It inherits the aliased type's hints and cannot declare its own." hint.at "hints are not allowed on a plain type definition"
```
jaehyun1ee: Relates to the comment in `frontend/error.ml`. If we don't care whether there is a bar or not and whether there are single/multi cases, then we might not need the `code` parameter here.

p4spec/bin/cli_error.ml
```diff
@@ -0,0 +1,16 @@
...
+
+let run_with_diagnostics ~(on_success : 'a -> unit)
+    (action : unit -> 'a P4spectec.result) : unit =
+  let result, report = P4spectec.with_diagnostics action in
+  render_diagnostics report;
+  match result with
+  | Ok value ->
+      let (), report = P4spectec.with_warnings (fun () -> on_success value) in
+      render_diagnostics report
+  | Error _ -> ()
```
jaehyun1ee: This I believe is a sequencing helper for unwrapping result types. For better readability, why not also make the `action` parameter named, and also reorder the params?

```
let run_with_diagnostics ~(do : unit -> 'a P4spectec.result) ~(on_success : 'a -> unit) : unit =
```

This way I think the (kind of) sequence order is preserved on `run_with_diagnostics` calls, so we can read the control flow from up-to-down order:

```
run_with_diagnostics
   ~do:(fun () -> ... res)
   ~on_success:(fun a -> ...)
```

p4spec/lib/lang/el/ast.ml
```diff
@@ -160,7 +160,7 @@ and prem' =
 
 (* Hints *)
 
-and hint = { hintid : id; hintexp : exp }
+and hint = { hintid : id; hintexp : exp; at : region }
```
jaehyun1ee: Would it make sense to define as follows, for coherence?

```ocaml
type hint = hint' phrase
and hint' = { hintid : id; hintexp : exp }
```

p4spec/lib/pass/algo/binding/iterctx.ml
```diff
@@ -70,12 +70,25 @@ let filter_bound (f : var -> bool) (iterctx : t) : t =
...
+            vars_bind
+        in
+        let cause =
+          match variables with
+          | [ variable ] -> variable ^ " is newly bound"
+          | _ ->
+              "variables " ^ String.concat ", " variables ^ " are newly bound"
+        in
+        error ~code:Iteration_missing_loop_variable at
+          ("iteration has no loop variable because " ^ cause)
```
jaehyun1ee: Minor, but "loop variable" → "iter variable"?

p4spec/lib/pass/algo/attempt.ml
```diff
@@ -10,4 +8,6 @@ let ( let* ) (attempt : 'a attempt) (f : 'a -> 'b) : 'b =
 let ( let+ ) (attempt : 'a attempt) (f : 'a -> 'b) : 'b =
   match attempt with
   | Ok a -> f a
-  | Fail failtraces -> error no_region (string_of_failtraces_short failtraces)
+  | Fail _ ->
+      (* Corresponding inputs in a relation's rules have equivalent types. *)
+      assert false
```
jaehyun1ee, line 13: Not sure if the comment above the `assert false` is correct?

p4spec/lib/pass/algo/binding/analyze.ml
```diff
@@ -111,9 +114,12 @@ let analyze_exps_as_bind (ctx : Ctx.t) (iterctx : Iterctx.t) (exps : exp list) :
...
   if not (BEnv.is_empty binds) then
-    error exp.at
-      (Format.asprintf "expression has free variable(s): %s"
-         (BEnv.to_string binds))
+    error ~code:Free_variable_in_expression exp.at
+      (Format.asprintf "expression uses unbound %s"
+         (BEnv.describe_variables binds))
+      ~detail:
+        "Every variable here must already be bound by an earlier part of the \
+         rule, such as the conclusion's input position or a preceding premise."
```
jaehyun1ee: "conclusion's input position" → "relation's input"?

p4spec/lib/pass/algo/binding/analyze.ml
```diff
@@ -476,19 +529,27 @@ let analyze_clause ?(is_else : bool = false) (ctx : Ctx.t) (clause : clause) :
...
 
 let analyze_elseclause (ctx : Ctx.t) (elseclause : elseclause) : elseclause =
   analyze_clause ~is_else:true ctx elseclause
 
 (* Table row binding analysis *)
 
+let error_pattern_type_without_cases code ?(related = []) (typ : typ) : 'a =
+  error ~code ~related typ.at
+    (Format.asprintf
+       "table row patterns require a type with declared cases, but got %s"
```
jaehyun1ee: "a type with declared cases" → "a variant type"?

p4spec/lib/pass/algo/error.ml
```diff
@@ -1,13 +1,57 @@
...
+  | Repeated_table_binding
+  | Relation_call_in_otherwise
+  | Function_call_in_otherwise
+  | Condition_in_otherwise
+  | Bind_both_sides_of_equality
+  | Non_invertible_binding
+  | Parallel_binding_dimension_mismatch
+  | Pattern_named_variant_expected
+  | Pattern_extern_type
+  | Pattern_non_variant_type
```
jaehyun1ee: Are they meaningfully different codes? I think `Pattern_non_variant_type` kind of subsumes the rest.

p4spec/lib/pass/algo/binding/ctx.ml
```diff
@@ -55,10 +55,17 @@ let add_bounds (ctx : t) (venv : VEnv.t) : t =
 
 (* Finders *)
 
-let find_typdef_opt (ctx : t) (tid : TId.t) : Typdef.t option =
-  TDEnv.find_opt tid ctx.tdenv
+let find_typdef_with_at (ctx : t) (tid : TId.t) : region * Typdef.t =
+  match
+    TDEnv.find_first_opt (fun key -> TId.compare key tid >= 0) ctx.tdenv
+  with
+  | Some (key, typdef) when TId.compare key tid = 0 -> (key.at, typdef)
```
jaehyun1ee: Why not just match on `TDEnv.find_first_opt (fun tid_key -> TId.compare tid_key tid = 0) ctx.tdenv`?

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1295,8 +1516,14 @@ and elab_prems (ctx : Ctx.t) (prems : prem list) : Ctx.t * prem_internal list =
 (* Elaboration of variable premises *)
 
 and elab_var_prem (ctx : Ctx.t) (id : id) (plaintyp : plaintyp) : Ctx.t =
-  check (valid_tid id) id.at "invalid meta-variable identifier";
-  check (not (Ctx.bound_typdef ctx id)) id.at "type already defined";
+  check ~code:Var_prem_invalid_metavar (valid_tid id) id.at
+    (F.asprintf "meta-variable identifier `%s` must not have a numeric suffix"
```
jaehyun1ee: "must not have a numeric suffix" → "must not have a suffix"? Only the existence of a suffix should matter, regardless of whether it is `x_1` or `x_a`.

Applies to everywhere else in `elab.ml`.

p4spec/lib/pass/elaborate/ctx.ml
```diff
@@ -242,123 +306,157 @@ let add_tparams (ctx : t) (tparams : tparam list) : t =
...
+  (* [elab_table_def_def] has already found [fid] as a table. *)
+  assert (find_table_func_opt ctx fid |> Option.is_some);
   let params, plaintyp, tablerows_found = find_table_func ctx fid in
   if List.length tablerows_found > 0 then
-    error_dup (List.hd tablerows_found).at "table function" fid.it;
+    error ~code:Table_rows_redefined
+      ~related:[ ((List.hd tablerows_found).at, "first row definition here") ]
+      fid.at
+      (Format.asprintf "table rows for function `%s` were already defined"
+         fid.it);
```
jaehyun1ee: This is a bit imprecise. We want to prevent `tbl def` from being redefined, so "table rows" can be misleading. Maybe put "table `%s` was already defined", instead?

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1319,9 +1546,20 @@ and elab_rule_not_prem (ctx : Ctx.t) (id : id) (exp : exp) : Ctx.t * Il.prem' =
...
+      error ~code:Negated_premise_has_outputs exp_output.at
+        (F.asprintf
+           "negated rule premise for relation `%s` cannot use output positions"
+           id.it)
+        ~related:
+          [ (nottyp_il.at, "relation signature with output positions here") ]
+        ~detail:
+          "A negated rule premise asserts that a relation does not hold for \
+           given inputs. Output positions have no value to produce when the \
+           relation fails."
```
jaehyun1ee: I'm not sure if the note is clear enough. Maybe simply put "Rule premise negation is supported only for relations without outputs. This constraint may be relaxed in future versions."

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1416,12 +1688,34 @@ let elab_clause (ctx : Ctx.t) (at : region) (id : id) (tparams : tparam list)
...
+    ~detail:
+      "A `def $f<...> = ...` clause must repeat the type parameters from its \
+       `dec` declaration with the same count and the same names in the same \
+       order.";
+  let args_expected = List.length params_il in
+  let args_actual = List.length args in
+  check ~code:Clause_arg_arity_mismatch
+    (args_expected = args_actual)
+    (region_of_first_extra args_expected (List.map (fun arg -> arg.at) args) at)
+    (F.asprintf "function `%s` expects %d argument%s, but this clause has %d"
```
jaehyun1ee: "argument%s" → "parameter%s"?

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -160,15 +221,65 @@ and elab_typcase_plain (ctx : Ctx.t) (typ_il : Il.typ) : Il.typcase list =
...
+            ~related:
+              (match Ctx.typdef_prior_at ctx tid with
+              | Some at -> [ (at, "extern type declared here") ]
+              | None -> [])
+            ~detail:
+              "`| T` extends a variant with `T`'s cases, but extern types have \
+               no cases.")
+  | _ ->
+      error ~code:Extend_non_variant_primitive typ_il.at
+        (F.asprintf "primitive type %s does not allow extension"
```
jaehyun1ee: Maybe put the other way, "extension is not allowed for [...]"? Applies elsewhere.

p4spec/lib/pass/elaborate/elab.ml
```diff
@@ -1224,25 +1382,85 @@ and elab_arg ?(as_def = false) (ctx : Ctx.t) (param_il : Il.param) (arg : arg) :
...
-           (Id.to_string id_p));
+           "passed function `%s` must have the same signature as function \
+            parameter `%s`"
+           (Id.to_string id_a) (Id.to_string id_p))
+        ~related
+        ~detail:
+          "A function argument at a call site must have the same signature \
+           (type parameters, parameter types, and return type) as the declared \
+           function parameter. The function passed here was declared with a \
+           different signature.";
```
jaehyun1ee: How about adding this detail note also for `Functyp_{t}param_arity_mismatch` as well?

jaehyun1ee:
Thank you for the substantial cleanup! Very happy to see polished error messages :) I think I'm halfway through the review, hoping to submit a review by tomorrow or Friday.

p4spec/test/neg/cases/annotate/annotation-type-parameter-shadows-global.watsup
jaehyun1ee, line 1: Not sure what feature this test is targeting? The `.expect` file also doesn't seem to warn, nor emit an error for this test case.

jaehyun1ee (changes requested):
I've reviewed all except for `p4spec/lib/diagnostic/*` and the test driver codes in `p4spec/test/*`. I'm sure they can use some polishing, but given their sheer size and complexity, seems not worth it for now.

I also made some nit passes myself where refactoring was quite evident, but left questions concerning the error message and some api design as comments.

After resolving these, I think I can resolve the merge conflicts and merge the PR.

Below is a coarse-grained summary of the commits I added:

- **Interpreter failures flow through backtracking.** The interpreters no longer raise `InterpError`/`BacktrackError`; runtime faults are `Err` failtraces like any other.

```diff
 (* interp-common/error.ml *)
-exception InterpError of region * string
-exception BacktrackError of Util.Attempt.failtrace list
-
-let error (at : region) (msg : string) = raise (InterpError (at, msg))
-let error_with_failtraces failtraces = raise (BacktrackError failtraces)
+let error (at : region) (msg : string) : error =
+  Diagnostic.error ~source:"interp" at msg
```

- **One failure vocabulary at the boundaries.** `Run.failure` is
  `Abort | Unmatch` and `Error.t` is `Diagnostic.t`. The diagnostic's `source` field replaces the per-stage error tags.

```diff
 (* runtime/dynamic-runner/signature.ml *)
-type failure =
-  | Diagnostic of Diagnostic.t
-  | Failtraces of Util.Attempt.failtrace list
+(* an abort propagates as is; an unmatch re-enters the caller's backtracking *)
+type failure = Abort of error | Unmatch of Util.Attempt.failtrace list
```

```diff
 (* lib/error.ml — the diagnostic's [source] replaces the stage tags *)
-type t =
-  | PassError of Pass.error
-  | RunError of Run.error
-  | CommandError of string
-  | CommandDiagnostic of Diagnostic.t
+type t = Diagnostic.t

 (* lib/p4spectec.ml *)
 let elab (paths_spec : string list) : Lang.Il.spec result =
-  Pass.elab paths_spec |> Result.map_error (fun e -> Error.PassError e)
+  Pass.elab paths_spec
```

- **Hints carry their own provenance.** Field/input hints keep element regions and `Annot.hint` pairs each hint with its id end to end, removing the parallel located/bare representations and the conversion layer between them.
```diff
 (* lang/pl/annot.ml — a hint keeps its id from HEnv to the renderer *)
+type 'a hint = { id : HId.t; value : 'a }
+
 type hints = {
-  prose : Hints.Alter.t option;
-  prose_in : Hints.Alter.t option;
+  prose : Hints.Alter.t hint option;
+  prose_in : Hints.Alter.t hint option;
   ...
-  prose_fields : Hints.Fields.t option;
+  prose_fields : Hints.Fields.t hint option;
 }

 (* pass/annotate/annotate.ml — the conversion layer is gone *)
-let hint_values (hints : located_hints) : Annot.hints =
-  { prose = Ctx.hint_value hints.prose; prose_in = Ctx.hint_value hints.prose_in; ... }
```

p4spec/lib/frontend/error.ml
KunJeong: I think I was leaning too hard on the one-test-per-errror-site rule. Consolidated to `Hint_on_plain_type` in 2b629fa6114f.

p4spec/lib/frontend/parser.mly
KunJeong: Addressed in 2b629fa6114f.

dune-project
KunJeong: Addressed in 755e429bee73.

p4spec/bin/cli_error.ml
KunJeong: I followed an OCaml convention that puts the main thunk last, but here I agree that your version is much more readable. Addressed in e0fed0583ebc.

p4spec/lib/pass/algo/attempt.ml
KunJeong: Comment attached incorrectly because `let+` was used in only one site. Fixed in 772b3757055c.

p4spec/lib/pass/algo/error.ml
KunJeong: Merged into `Pattern_non_variant_type` in 2b629fa6114f.

p4spec/lib/pass/algo/binding/ctx.ml
KunJeong: It's ugly, but necessary because `find_first_opt f m` requires `f` to be a monotonically increasing function ([OCaml reference](https://ocaml.org/manual/5.1/api/Map.S.html#VALfind_first_opt)). If you prefer, I could extract this into a helper function.

p4spec/lib/lang/el/ast.ml
KunJeong: Addressed in e0fed0583ebc.

p4spec/lib/frontend/parser.mly
KunJeong: Addressed in 7e2693207200.

p4spec/lib/pass/algo/binding/iterctx.ml
KunJeong: Addressed in 7e2693207200.

p4spec/lib/pass/algo/binding/analyze.ml
KunJeong: Addressed in 7e2693207200.

p4spec/lib/pass/elaborate/elab.ml
KunJeong: Addressed in 7e2693207200.

p4spec/lib/pass/elaborate/elab.ml
KunJeong: Addressed in 7e2693207200, but I dropped the "constraint may be relaxed in future versions"  note.

p4spec/lib/pass/elaborate/elab.ml
KunJeong: Addressed in 7e2693207200.

p4spec/lib/pass/elaborate/elab.ml
KunJeong: Addressed in 7e2693207200.

p4spec/lib/pass/elaborate/elab.ml
KunJeong: Reworded to `` function `%s` was declared with %d parameter%s, but this clause has %d `` in 7e2693207200.

p4spec/lib/pass/algo/binding/analyze.ml
KunJeong: Addressed in 7e2693207200.

p4spec/lib/pass/elaborate/ctx.ml
KunJeong: Addressed in 7e2693207200.

p4spec/test/neg/cases/annotate/annotation-type-parameter-shadows-global.watsup
KunJeong: This pinned an incidental annotation fix, which I should have explained further.

```
extern dec $f<T>(T) : T
syntax T = nat
```

Elaboration accepts this because `$f` introduces its own type parameter `T` before the global `T` is declared.
Annotation used to reject it with “type `T` was already defined”. It collected all global declarations first, which failed with duplicate names when adding $f’s local parameter. The fix allows that parameter to replace `T` in the function’s local context.

Since it's rather confusing alongside the diagnostic work, I've removed the test in 44c915b2fda5.

KunJeong:
Thanks for the cleanup, quick review, and great suggestions on diagnostics. I've applied most of the feedback, two of the diagnostic suggestions were reworded slightly and two comments need another look. Please take a look and let me know 😃

p4spec/lib/pass/elaborate/elab.ml
jaehyun1ee: Actually, I want to retain the phrase since I'm still quite open to the relaxation. Introduction of rule negation (only for output-less relations) was an ad-hoc decision at the first place, and I don't want it to be rendered as a hard constraint.

KunJeong: Added in f1e963928f3.

KunJeong: Sorry, missed the .expected update. Amended to b42f26df6033.

KunJeong:
All comments addressed.

jaehyun1ee (approved):
Thanks again for the effort, merging!

@pacokwon might want to have a look, considering the upcoming tutorial.

PR #301, opened 2026-09-16
Fold deep trace indentation and exit non-zero on failed runs

jaehyun1ee:
## Summary

Four changes to how the CLI reports a failed run, found while reading one long `Program_ok` trace. The trace no longer walks off the terminal, a failed `cc -E` is an error instead of an empty program that passes, and a failed `run`/`sim` exits with 1. One dead-code removal rides along.

**1. Fold deep trace indentation.** `Diagnostic.Render.render_trace_node` grew the indent four columns per level, so a `Decls_ok` chain over the ~30 declarations of `core.p4` ran past 140 columns. Every `fold_depth = 4` levels below the last anchor, a chain restarts at the trace base after a dim cue line. `const int<0> x = 1;` with `core.p4` now renders as:

```
  trace:
  spec/5-typing/5.11-typing-declaration.watsup:1168.6-1168.14
  relation Decls_ok failed
  └── spec/5-typing/5.11-typing-declaration.watsup:1155.8-1155.16
      relation Decls_ok failed
      └── ...
              └── spec/5-typing/5.11-typing-declaration.watsup:1155.8-1155.16
                  relation Decls_ok failed
  ⋮ (depth 5, continued)
  └── spec/5-typing/5.11-typing-declaration.watsup:1155.8-1155.16
      relation Decls_ok failed
      └── ...
  ⋮ (depth 9, continued)
  └── spec/5-typing/5.11-typing-declaration.watsup:1153.8-1153.15
      relation Decl_ok failed
      └── ... ConstDecl_ok ... Type_ok ...
              └── spec/5-typing/5.05.1-typing-type.watsup:46.11-46.19
                  condition (n > 0) was not met
```

A restart only happens when no sibling bar is pending below the anchor (every ancestor since the anchor was the last child), so a branch never loses its `│` line. `Record.trace_node` is unchanged; this is render-only.

```ocaml
type layout = {
  base : string;      (* trace base indent *)
  indent : string;    (* this node's connector column *)
  level : int;        (* absolute depth *)
  anchor : int;       (* depth whose children sit at [base] *)
  foldable : bool;    (* no sibling bar pending below [anchor] *)
}
```

**2. Report a failed preprocessor run.** `Preprocessor.preprocess` dropped the exit status of `cc -E`. A missing header, or a missing input file, yielded an empty program that parsed and type-checked as `passed`:

```diff
-  let _ = Core_unix.close_process_in in_chan in
-  program
+  match Core_unix.close_process_in in_chan with
+  | Ok () -> Ok program
+  | Error _ as status ->
+      Error ("cc " ^ Core_unix.Exit_or_signal.to_string_hum status)
```

`Parse.preprocess` raises `ParseError` at the file's region, so the diagnostic carries a `-->` location; `cc`'s own stderr still passes through with the header name.

```
error: preprocessing failed: cc exited with code 1
  --> prog.p4
  |
  | source: p4
```

**3. Exit with 1 when a run fails.** `Cli_error.run_with_diagnostics` only exited non-zero when the `action` stage failed. A `Fail` inside `on_success` was rendered and the process exited 0, so every ill-typed program and every preprocessing failure looked like success to a shell. `on_success` now returns `unit P4spectec.result`; the driver renders the error together with any warnings and exits:

```diff
 let run_with_diagnostics ~(action : unit -> 'a P4spectec.result)
-    ~(on_success : 'a -> unit) : unit =
+    ~(on_success : 'a -> unit P4spectec.result) : unit =
   let result, report = P4spectec.with_diagnostics action in
   render_diagnostics report;
   match result with
+  | Error _ -> exit 1
   | Ok value ->
-      let (), report = P4spectec.with_warnings (fun () -> on_success value) in
-      render_diagnostics report
-  | Error _ -> exit 1
+      let result, report =
+        P4spectec.with_diagnostics (fun () -> on_success value)
+      in
+      render_diagnostics report;
+      if Result.is_error result then exit 1
```

`run`, `sim`, boot `run`/`boot-n` and both `parse` commands return `Error diagnostic` instead of calling `render_diagnostics` themselves; `parse`'s `Sys_error` fallback, which printed `File error:` to stdout and exited 0, becomes a diagnostic. `interesting` keeps its own numeric exit codes, which the fuzzer depends on. `fuzzer/gauntlet.py` classifies by stdout and ignores the return code; CI and the Makefile do not consume `run`/`sim` exit codes.

**4. Drop dead code in `Util.Attempt`.** `short_window`, `region_line`, `string_of_failtrace(s)` and `string_of_failtraces_short` had no callers since traces render through `Diagnostic.Render`. `depth_of` stays for `Backtrace`.

## Validation

- `dune build @speclang` passes, so `neg/interp.expected` and the other goldens are unchanged: the deepest existing golden trace is 3 levels, below `fold_depth`.
- Manual, with `main.exe run spec -rel Program_ok ... -sl`: `p4c/testdata/p4_16_samples/arith-bmv2.p4` prints `passed` and exits 0; `const int<0> x = 1;` renders the folded trace above and exits 1; `#include <no-such-header.p4>` and a nonexistent `-p` path report `preprocessing failed` and exit 1 where they previously printed `passed`.
- `dune build @fmt` is clean for every touched file.
- No tests were added. Not run: `make test-fast`, `make test-all`. The `p4parse`/`run-sl`/`sim-sl` suites go through the library, not the CLI, and the p4c submodule was not checked out in the worktree.


jaehyun1ee:
@pacokwon PTAL!

